PR #26, opened 2017-01-31
Runtime configuration support

### LINKED ISSUE #25: Missing support for a runtime configuration file
The runtime needs a configuration file for runtime specific settings:

- Hypervisor
  - hypervisor path
  - Guest kernel path
  - Guest OS image path

- Agent
  - Agent configuration (socket paths)

sameo:
We configure the runtime through a simple TOML
file that describes the hypervisor and proxy settings.

Fixes #25


config.go
```diff
@@ -0,0 +1,132 @@
...
+)
+
+const (
+	defaultRuntimeLib           = "/var/lib/cc-runtime"
+	defaultRuntimeRun           = "/var/run/cc-runtime"
+	defaultruntimeConfiguration = "/etc/cc-runtime/configuration.toml"
+	defaultHypervisor           = vc.QemuHypervisor
+	defaultProxy                = vc.CCProxyType
+	defaultAgent                = vc.HyperstartAgent
+)
```
sboeuf: Maybe we want a default hypervisor config too ?

config.go
```diff
@@ -0,0 +1,132 @@
...
+			break
+		}
+	}
+
+	for k, agent := range tomlConf.Agent {
+		switch k {
+		case hyperstart:
+			aConfig = vc.HyperConfig{
+				SockCtlName: agent.Ctl,
+				SockTtyName: agent.IO,
```
sboeuf: I would prefer there is no way for us to specify those sockets as they could be not unique. We already have a mechanism in agent implementation for Hyperstart which is creating some unique socket names based on the podID.

sboeuf:
@sameo I think we should also add the proxy socket path in this configuration file. So that the runtime has this information and it can provide it to the shim. Indeed, the shim needs to know how to connect to the proxy.

config.go
```diff
@@ -0,0 +1,163 @@
...
+	"github.com/BurntSushi/toml"
+	log "github.com/Sirupsen/logrus"
+	vc "github.com/containers/virtcontainers"
+	"github.com/containers/virtcontainers/pkg/oci"
+)
+
+const (
+	defaultRuntimeLib           = "/var/lib/cc-runtime"
+	defaultRuntimeRun           = "/var/run/cc-runtime"
+	defaultruntimeConfiguration = "/etc/cc-runtime/configuration.toml"
```
sboeuf: Don't we want defaultRuntimeConfiguration name instead ?

config.go
```diff
@@ -0,0 +1,163 @@
...
+
+	config := oci.RuntimeConfig{
+		HypervisorType:   defaultHypervisor,
+		HypervisorConfig: defaultHypervisorConfig,
+		AgentType:        defaultAgent,
+		ProxyType:        defaultProxy,
+	}
+
+	if configPath == "" {
+		configPath = defaultruntimeConfiguration
```
sboeuf: Ditto

config.go
```diff
@@ -0,0 +1,163 @@
...
+	}
+
+	log.Debugf("TOML configuration: %v", tomlConf)
+
+	if len(tomlConf.Hypervisor) > 1 {
+		return config, errTooManyHypervisors
+	}
+
+	if len(tomlConf.Proxy) > 1 {
+		return config, errTooManyAgents
```
sboeuf: Maybe we should try to be more consistent about the naming and stick to Agent or Proxy, but avoid mixing to avoid confusion.

config.go
```diff
@@ -0,0 +1,163 @@
...
+	}
+
+	for k, hypervisor := range tomlConf.Hypervisor {
```
sboeuf: We shouldn't use for{} loop as we know at that point there is only one Hypervisor from the configuration file. It would avoid unnecessary indenting.

config.go
```diff
@@ -0,0 +1,163 @@
...
+				KernelPath:     hypervisor.kernel(),
+				ImagePath:      hypervisor.image(),
+			}
+
+			config.HypervisorConfig = hConfig
+			break
+		}
+	}
+
+	for k, proxy := range tomlConf.Proxy {
```
sboeuf: Ditto

config_test.go
```diff
@@ -0,0 +1,163 @@
...
+	if reflect.DeepEqual(config, expectedConfig) == false {
+		t.Fatalf("Got %v\n expecting %v", config, expectedConfig)
+	}
+
+	if err := os.Remove(configPath); err != nil {
+		t.Fatal(err)
+	}
+}
+
+func TestMinimalRuntimeConfig(t *testing.T) {
```
sboeuf: You could factorize those two unit tests just giving runtimeConfig or runtimeMinimalConfig as a parameter.

config.go
sameo: Unfortunately this is a map, and we can't just take the "first" entry.

sboeuf:
LGTM

config.go
```diff
@@ -0,0 +1,163 @@
...
+	"github.com/BurntSushi/toml"
+	log "github.com/Sirupsen/logrus"
+	vc "github.com/containers/virtcontainers"
+	"github.com/containers/virtcontainers/pkg/oci"
+)
+
+const (
+	defaultRuntimeLib           = "/var/lib/cc-runtime"
+	defaultRuntimeRun           = "/var/run/cc-runtime"
+	defaultRuntimeConfiguration = "/etc/cc-runtime/configuration.toml"
```
dlespiau: I was thinking that we should stick everything (runtime/proxy/shim?) config files under the same directories, say `clearcontainers` for 3.0:

```
/etc/clearcontainers/configuration.tom
/var/run/clearcontainers/proxy.sock
/var/run/clearcontainers/proxy-shim.sock
/var/run/clearcontainers/runtime/
```

(can be done at a later time if at all, of course)

config.go
```diff
@@ -0,0 +1,163 @@
...
+	defaultHypervisor           = vc.QemuHypervisor
+	defaultProxy                = vc.CCProxyType
+	defaultAgent                = vc.HyperstartAgent
+	defaultKernelPath           = "/usr/share/clear-containers/vmlinux.container"
```
dlespiau: and so here clearcontainers as well (or clear-containers or whatever)

dlespiau:
lgtm

PR #27, opened 2017-01-31
Implement OCI runtime specification

sboeuf:
This PR implements the create function relying on OCI runtime specification.

### DIFF
--- a/create.go
+++ b/create.go
@@ -16,6 +16,11 @@
 package main
 
 import (
+	"fmt"
+	"os"
+
+	vc "github.com/containers/virtcontainers"
+	"github.com/containers/virtcontainers/pkg/oci"
 	"github.com/urfave/cli"
 )
 
@@ -52,7 +57,84 @@ var createCommand = cli.Command{
 		},
 	},
 	Action: func(context *cli.Context) error {
-		// TODO
-		return nil
+		return create(context.String("container-id"),
+			context.String("bundle"),
+			context.String("console"),
+			context.String("pid-file"))
 	},
+}
+
+func create(containerID, bundlePath, console, pidFilePath string) error {
+	// Checks the MUST and MUST NOT from OCI runtime specification
+	if err := validCreateParams(containerID, bundlePath); err != nil {
+		return err
+	}
+
+	podConfig, err := getPodConfig(bundlePath, containerID, console)
+	if err != nil {
+		return err
+	}
+
+	pod, err := vc.CreatePod(podConfig)
+	if err != nil {
+		return err
+	}
+
+	// Start the shim to retrieve its PID.
+	pid, err := startShim(pod)
+	if err != nil {
+		return err
+	}
+
+	// Creation of PID file has to be the last thing done in the create
+	// because containerd considers the create complete after this file
+	// is created.
+	if err := createPIDFile(pidFilePath, pid); err != nil {
+		return err
+	}
+
+	return nil
+}
+
+func getPodConfig(bundlePath, containerID, console string) (vc.PodConfig, error) {
+	runtimeConfig, err := loadConfiguration("")
+	if err != nil {
+		return vc.PodConfig{}, err
+	}
+
+	podConfig, err := oci.PodConfig(runtimeConfig, bundlePath, containerID, console)
+	if err != nil {
+		return vc.PodConfig{}, err
+	}
+
+	return *podConfig, nil
+}
+
+func createPIDFile(pidFilePath string, pid int) error {
+	if pidFilePath == "" {
+		return fmt.Errorf("Missing PID file path")
+	}
+
+	if err := os.RemoveAll(pidFilePath); err != nil {
+		return err
+	}
+
+	f, err := os.Create(pidFilePath)
+	if err != nil {
+		return err
+	}
+	defer f.Close()
+
+	pidStr := fmt.Sprintf("%d", pid)
+
+	n, err := f.WriteString(pidStr)
+	if err != nil {
+		return err
+	}
+
+	if n < len(pidStr) {
+		return fmt.Errorf("Could not write pid to '%s': only %d bytes written out of %d", pidFilePath, n, len(pidStr))
+	}
+
+	return nil
 }
--- a/delete.go
+++ b/delete.go
@@ -16,6 +16,10 @@
 package main
 
 import (
+	"fmt"
+
+	vc "github.com/containers/virtcontainers"
+	"github.com/containers/virtcontainers/pkg/oci"
 	"github.com/urfave/cli"
 )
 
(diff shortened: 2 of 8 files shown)


create.go
```diff
@@ -52,7 +57,115 @@ var createCommand = cli.Command{
...
+		return fmt.Errorf("Missing bundle path")
+	}
+
+	// bundle path MUST be valid.
+	fileInfo, err := os.Stat(bundlePath)
+	if err != nil {
+		return fmt.Errorf("Invalid bundle path: %s", err)
+	}
+	if fileInfo.IsDir() == false {
+		return fmt.Errorf("Invalid bundle path, it should be a directory")
```
jodh-intel: It would be useful to add `bundlePath` to the error.

create.go
```diff
@@ -52,7 +57,115 @@ var createCommand = cli.Command{
...
+
+	pidStr := fmt.Sprintf("%d", pid)
+
+	n, err := f.WriteString(pidStr)
+	if err != nil {
+		return err
+	}
+
+	if n < len(pidStr) {
+		return fmt.Errorf("Could write pid to pid file: only %d bytes written out of %d", n, len(pidStr))
```
jodh-intel: Add `pidFilePath` to error?

create.go
sboeuf: Good idea :)

create.go
sboeuf: Good idea :)

sboeuf:
@dlespiau, @sameo please take a look now that is building properly.

create.go
```diff
@@ -52,7 +57,124 @@ var createCommand = cli.Command{
...
+
+	if err := os.RemoveAll(pidFilePath); err != nil {
+		return err
+	}
+
+	f, err := os.Create(pidFilePath)
+	if err != nil {
+		return err
+	}
+
```
sameo: `defer f.Close()`

shim.go
```diff
@@ -0,0 +1,88 @@
...
+type ShimInfo struct {
+	Path  string
+	Token string
+	IP    string
+	Port  string
+}
+
+// Specific values related to the shim.
+const (
+	defaultShimPath  = "/usr/local/libexec/cc-shim"
```
sameo: Can we put that one in `config.go` ?

shim.go
```diff
@@ -0,0 +1,88 @@
...
+	defaultProxyIP   = "192.168.1.1"
```
sameo: Default IP should be `127.0.0.1`

shim.go
sameo: Also, as a follow up patch, the shim configuration (That should include the path, IP and port) should be added to the config.go `tomlConfig` structure.

shim.go
```diff
@@ -0,0 +1,88 @@
...
+}
+
+func startShim(shimInfo ShimInfo) (int, error) {
+	attr := os.ProcAttr{
+		Env: os.Environ(),
+	}
+
+	args := []string{"-t", shimInfo.Token, "-s", shimInfo.IP, "-p", shimInfo.Port}
+
+	process, err := os.StartProcess(shimInfo.Path, args, &attr)
```
sameo: Could we please use the `exec` package here ? `Cmd` and `Cmd.Start()`.

create.go
```diff
@@ -52,7 +57,124 @@ var createCommand = cli.Command{
...
+	// because containerd considers the create complete after this file
+	// is created.
+	if err := createPIDFile(pidFilePath, pid); err != nil {
+		return err
+	}
+
+	return nil
+}
+
+func checkCreateParamsIntegrity(containerID, bundlePath string) error {
```
sameo: Could we have an `oci.go` file where we manage the spec compliance parts ?
So all the parameters validation (And not integrity ;)) routines should be added there.

Also please not that except for `create`, all other commands in this PR only validate the container ID. So having a `validContainer(containerID, command string) (bool, error)` generic routine would make the code overall more readable.

create.go
```diff
@@ -52,7 +57,124 @@ var createCommand = cli.Command{
...
+	// Start the shim to retrieve its PID.
+	pid, err := startShim(shimInfo)
+	if err != nil {
+		return err
+	}
+
+	if err := storeShimPid(pod, pid); err != nil {
+		return err
+	}
+
```
sameo: Is there any case where we're going to start a shim without storing its PID ? it seems to me this is an atomic operation and you could have `startShim(pod)` do `getShimInfo()`, then start the shim process and finally `storeShimPid()`. `startShim(pod)` would return the shim PID.

delete.go
```diff
@@ -39,7 +42,41 @@ EXAMPLE:
...
+func delete(containerID string, force bool) error {
+	// Checks the MUST and MUST NOT from OCI runtime specification
+	if err := checkDeleteParamsIntegrity(containerID); err != nil {
+		return err
+	}
+
+	if _, err := vc.StopPod(containerID); err != nil {
+		return err
+	}
+
```
sameo: We have to handle `--force` here. If the container is running and `--force` is not set, we should return an error otherwise we stop the pod and delete.

state.go
```diff
@@ -28,7 +34,81 @@ var stateCommand = cli.Command{
...
+
+	stateJSON, err := json.Marshal(state)
+	if err != nil {
+		return "", err
+	}
+
+	return string(stateJSON), nil
+}
+
+func stateToOCIState(state vc.State) string {
```
sameo: Ditto

state.go
```diff
@@ -28,7 +34,81 @@ var stateCommand = cli.Command{
...
+		return err
+	}
+	if exist == false {
+		return fmt.Errorf("Container ID does not exist")
+	}
+
+	return nil
+}
+
+func statusToStateJSON(status vc.PodStatus) (string, error) {
```
sameo: Can you please move that one to the virtcontainers OCI package ?

create.go
sboeuf: Done

create.go
sboeuf: Done

shim.go
sboeuf: Done

shim.go
sboeuf: Done

delete.go
sboeuf: I agree with the idea, but we cannot do that. In order to do that, you would need to get the status switched from running to stopped inside virtcontainers, but in case we are calling KillContainer(), we don't update the pod/container state because we don't know what the signal did to the process. And because we rely on virtcontainers to get the state, there is no way a kill from OCI runtime will switch the state to stopped. Then if we don't force, we will always get an error...

sboeuf: I guess we can do the StopPod only if we force. And do only the Delete in case force == false.

shim.go
sboeuf: I agree about the follow up patch, but let's do that after we get this PR merged.

delete.go
sboeuf: @sameo I think I have a solution to know if the container is running. We just need to check if the shim PID still exists, right ?
In case we can find it, this means it is not stopped and we can return an error in case we don't have "force" enabled.
How does that sound ?

sameo: @sboeuf Yes, that's the runc approach as well. We need to actually update our container status file as well, through a virtcontainers `UpdateStatus()` container API. It should be called for our `List` APIs too.
Please do that as a follow up PR between runtime and virtcontainers.

oci.go
```diff
@@ -67,3 +67,35 @@ func validCreateParams(containerID, bundlePath string) error {
...
+		return err
+	}
+	if exist == false {
+		return fmt.Errorf("Container ID does not exist")
+	}
+
+	return nil
+}
+
+func processRunning(pid int) (bool, error) {
```
sameo: The more idiomatic way to search for a process is to send it signal 0. Something like:

```Go
process, err := os.FindProcess(int(pid))
if err != nil {
        return err
} else {
        if err := process.Signal(syscall.Signal(0)) ; err != nil {
                    return err
        }
            
        return nil
}
```

sboeuf: I didn't know about that, let me fix that.

sameo:
LGTM

sboeuf:
LGTM

PR #45, opened 2017-02-09
vendor: virtcontainers vendoring update

sboeuf:
In order to implement the OCI runtime relying on virtcontainers, we need
to rely on latest changes:
- Pod and containers state are now public to perform the translation to
OCI specific states.
- PID and RootFs stored inside the Pod object are copied to the PodStatus,
making those info available from "state" command using StatusPod() API.


dlespiau:
still looks fine :)

lgtm

PR #47, opened 2017-02-13
Update container's state if stopped

sboeuf:
The runtime implementation has to check if the process running inside the
container is still running. In case it is not running anymore, we want to
call StopContainer API from virtcontainers so that it removes the container
from the pod, and it gets the container status updated.

### DIFF
--- a/kill.go
+++ b/kill.go
@@ -91,17 +91,31 @@ func kill(containerID, signal string) error {
 		return err
 	}
 
-	// container MUST be running
 	podStatus, err := vc.StatusPod(containerID)
 	if err != nil {
 		return err
 	}
-	if podStatus.State.State != vc.StateRunning {
-		return fmt.Errorf("Container is not running")
-	}
 
 	if len(podStatus.ContainersStatus) != 1 {
-		return fmt.Errorf("ContainerStatus list from PodStatus is wrong, expecting only one ContainerStatus")
+		return fmt.Errorf("ContainerStatus list from PodStatus %s is wrong, expecting only one ContainerStatus", containerID)
+	}
+
+	// container MUST be running
+	if podStatus.ContainersStatus[0].State.State != vc.StateRunning {
+		return fmt.Errorf("Container %s is not running", containerID)
+	}
+
+	// Check status of process
+	running, err := processRunning(podStatus.ContainersStatus[0].PID)
+	if err != nil {
+		return err
+	}
+	if running == false {
+		if err := stopContainer(podStatus); err != nil {
+			return err
+		}
+
+		return fmt.Errorf("Process not running inside container %s", containerID)
 	}
 
 	if err := vc.KillContainer(containerID, podStatus.ContainersStatus[0].ID, signals[signal]); err != nil {
--- a/oci.go
+++ b/oci.go
@@ -99,3 +99,18 @@ func processRunning(pid int) (bool, error) {
 
 	return true, nil
 }
+
+func stopContainer(podStatus vc.PodStatus) error {
+	if len(podStatus.ContainersStatus) != 1 {
+		return fmt.Errorf("ContainerStatus list from PodStatus is wrong, expecting only one ContainerStatus")
+	}
+
+	// Calling StopContainer allows to make sure the container is properly
+	// stopped and removed from the pod. That way, the container's state is
+	// updated to the expected "stopped" state.
+	if _, err := vc.StopContainer(podStatus.ID, podStatus.ContainersStatus[0].ID); err != nil {
+		return err
+	}
+
+	return nil
+}
--- a/state.go
+++ b/state.go
@@ -61,6 +61,10 @@ func state(containerID string) error {
 		return err
 	}
 	if running == false {
+		if err := stopContainer(podStatus); err != nil {
+			return err
+		}
+
 		state.Status = "stopped"
 	}
 


kill.go
```diff
@@ -91,19 +91,33 @@ func kill(containerID, signal string) error {
...
-		return fmt.Errorf("Container is not running")
-	}
 
 	if len(podStatus.ContainersStatus) != 1 {
 		return fmt.Errorf("ContainerStatus list from PodStatus is wrong, expecting only one ContainerStatus")
 	}
 
+	// container MUST be running
+	if podStatus.ContainersStatus[0].State.State != vc.StateRunning {
+		return fmt.Errorf("Container is not running")
```
jodh-intel: Which container isn't running?

kill.go
```diff
@@ -91,19 +91,33 @@ func kill(containerID, signal string) error {
...
+	running, err := processRunning(podStatus.ContainersStatus[0].PID)
+	if err != nil {
+		return err
+	}
+	if running == false {
+		if err := updateStoppedContainer(podStatus); err != nil {
+			return err
+		}
+
+		return fmt.Errorf("Container is not running")
```
jodh-intel: Same comment as above.

kill.go
```diff
@@ -91,19 +91,33 @@ func kill(containerID, signal string) error {
...
 	podStatus, err := vc.StatusPod(containerID)
 	if err != nil {
 		return err
 	}
-	if podStatus.State.State != vc.StateRunning {
-		return fmt.Errorf("Container is not running")
-	}
 
 	if len(podStatus.ContainersStatus) != 1 {
 		return fmt.Errorf("ContainerStatus list from PodStatus is wrong, expecting only one ContainerStatus")
```
jodh-intel: Include `podStatus.ID` here?

kill.go
sboeuf: done

kill.go
sboeuf: done

kill.go
sboeuf: done

kill.go
```diff
@@ -91,17 +91,31 @@ func kill(containerID, signal string) error {
...
+	running, err := processRunning(podStatus.ContainersStatus[0].PID)
+	if err != nil {
+		return err
+	}
+	if running == false {
+		if err := updateStoppedContainer(podStatus); err != nil {
+			return err
+		}
+
+		return fmt.Errorf("Container %s is not running", containerID)
```
jodh-intel: I'm finding this chunk rather difficult to understand - should this message be "Process not running inside container" (because you previously checked that the container was running)?

oci.go
```diff
@@ -99,3 +99,18 @@ func processRunning(pid int) (bool, error) {
 
 	return true, nil
 }
+
+func updateStoppedContainer(podStatus vc.PodStatus) error {
```
jodh-intel: I don't understand the naming of this function - it seems to stop the container, in which case why not just call it `stopContainer`?

kill.go
sboeuf: Oh yes you're right , it would be clearer ;)

oci.go
sboeuf: I know it is a bit confusing because the specification forces us to do something confusing here... It updates the container state in case we detected the process is not running anymore.

kill.go
sboeuf: done

oci.go
sboeuf: done

sameo:
LGTM

sameo:
Waiting to get @jodh-intel's LGTM before merging

kill.go
```diff
@@ -91,17 +91,31 @@ func kill(containerID, signal string) error {
...
+	running, err := processRunning(podStatus.ContainersStatus[0].PID)
+	if err != nil {
+		return err
+	}
+	if running == false {
+		if err := stopContainer(podStatus); err != nil {
+			return err
+		}
+
+		return fmt.Errorf("Process not running inside container %s", containerID)
```
jodh-intel: This now looks clearer to me now, but looking at this afresh, the code looks racy: wouldn't it be safer to simply have `kill()` call `vc.StopContainer()` since that will/should handle checks for the pod, container and process?

sboeuf: What do you think is racy here ? And about the checks we're doing here, I think we have to do them at that level to ensure we are OCI compliant. Indeed, we can perform some checks from virtcontainers library but it is not OCI compliant, that's why we have to do them in the runtime implementation.

jodh-intel: well, what if you call `processRunning()`, it returns true, then the process dies before you call `stopContainer()`. Will `stopContainer()` complain "process not running"? If so, we could just call `stopContainer()` immediately.

sboeuf: In that case, we would not go inside the if statement and the vc.KillContainer() would maybe return an error. But I don't want to call stopContainer() all the time, the idea being to update the container status in case the process is not running anymore. A kill means sending a signal, not stopping the container.
Unless I don't get what you are suggesting, I think that calling stopContainer all the time would modify the behavior I am expecting with this patch.

sboeuf:
@jodh-intel any updates on this one ? I would say it worth to merge this now and check it is fully functional and that we are not hitting any race issue when the runtime will be functional.

jodh-intel:
@sboeuf - ok. I see that it is somewhat problematic if you want to conditionally call `stopContainer()`. I think we could see the occasional error from `stopContainer()` if the process dies, but if @sameo is still happy,

lgtm

PR #48, opened 2017-02-14
[DEPRECATED] Add ShimConfig to be loaded from toml config file

sboeuf:
In order to be consistent with the way we retrieve different configurations
related to the runtime, let's add a ShimConfig structure describing a config
for the shim. This configuration will be stored into the toml file where all
other configurations are stored and will be loaded at "create" time.

### DIFF
--- a/config.go
+++ b/config.go
@@ -34,9 +34,9 @@ const (
 	defaultKernelPath           = "/usr/share/clear-containers/vmlinux.container"
 	defaultImagePath            = "/usr/share/clear-containers/clear-containers.img"
 	defaultHypervisorPath       = "/usr/bin/qemu-lite-system-x86_64"
-	defaultShimPath             = "/usr/local/libexec/cc-shim"
-	defaultProxyIP              = "127.0.0.1"
-	defaultProxyPort            = "54321"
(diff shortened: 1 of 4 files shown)


config.go
```diff
@@ -48,16 +48,22 @@ const (
 	ccProxy = "cc"
 )
 
+const (
+	ccShim = "cc"
```
jodh-intel: If this is the name of the config file section for the shim, wouldn't it be better to call it `shim`?

jodh-intel: Oh - reading on, I see the test that shows `shim.cc`. Now I'm really confused - rather than using top-level entities `shim.cc` and `proxy.cc`, wouldn't it be better to create some sort of `cc` namespace and then define `shim` and `proxy` inside that? Can that be done with toml?

sboeuf:
@jodh-intel this PR will be abandoned because we decided to get those information (IP and port) from the proxy at the moment we connect to it.
@sameo is that correct ? In case it is, please confirm I can drop the PR.

jodh-intel:
@sboeuf - ack. I've added the label as a pullapprove safety-check.

PR #53, opened 2017-03-04
exec: Implement OCI CLI spec for exec

sboeuf:
This PR adds a first implementation of the exec command.

### DIFF
--- a/create.go
+++ b/create.go
@@ -81,7 +81,12 @@ func create(containerID, bundlePath, console, pidFilePath string) error {
 	}
 
 	// Start the shim to retrieve its PID.
-	pid, err := startShim(pod)
+	containers := pod.GetContainers()
+	if len(containers) != 1 {
+		return fmt.Errorf("BUG: Container list from pod is wrong, expecting only one container, found %d containers", len(containers))
+	}
+
+	pid, err := startContainerShim(containers[0], pod.URL())
 	if err != nil {
 		return err
 	}
--- a/exec.go
+++ b/exec.go
@@ -16,8 +16,25 @@
 package main
 
 import (
+	"encoding/json"
+	"fmt"
+	"io/ioutil"
+
+	vc "github.com/containers/virtcontainers"
+	"github.com/containers/virtcontainers/pkg/oci"
+	specs "github.com/opencontainers/runtime-spec/specs-go"
 	"github.com/urfave/cli"
 )
+
+type execParams struct {
+	ociProcess   specs.Process
+	cID          string
+	pidFile      string
+	console      string
+	detach       bool
+	processLabel string
+	noSubreaper  bool
+}
 
 var execCommand = cli.Command{
 	Name:  "exec",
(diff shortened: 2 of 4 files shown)


exec.go
```diff
@@ -91,7 +111,83 @@ EXAMPLE:
...
+		return err
+	}
+
+	podStatus, err := vc.StatusPod(params.cID)
+	if err != nil {
+		return err
+	}
+
+	if len(podStatus.ContainersStatus) != 1 {
+		return fmt.Errorf("ContainerStatus list from PodStatus %s is wrong, expecting only one ContainerStatus", containerID)
```
jodh-intel: Show statuses on error? Also, this feels like it might need a `BUG` prefix as this situation indicates a programming error rather than user error / resource availability issue I think?

sboeuf: Yes I agree, I will do the changes.

shim.go
```diff
@@ -30,7 +30,32 @@ type ShimInfo struct {
...
+		Env:  os.Environ(),
+	}
+
+	if err := cmd.Start(); err != nil {
+		return -1, err
+	}
+
+	return cmd.Process.Pid, nil
+}
+
```
sameo: Ok, I think it would be cleaner if we could have a common function for both cases: ` func startShim(proxy vc.CCProxyConfig, process *vc.Process) (int, error)`

And then have a `startContainerShim(proxy vc.CCProxyConfig, container *vc.Container)` that would convert a `vc.Container` into its init `vc.Process`.

This means our `ShimInfo` structure and `getShimInfo()` routine can be removed.

oci.go
```diff
@@ -17,11 +17,41 @@ package main
 import (
 	"fmt"
 	"os"
+	"strings"
 	"syscall"
 
 	vc "github.com/containers/virtcontainers"
 )
 
+func envsToEnvVars(envs []string) ([]vc.EnvVar, error) {
```
sboeuf: Of course, it makes more sense. I did that to avoid the re-vendoring of virtcontainers, but now that we have this PR#44 merged in virtcontainers, I will rely onto it :)

shim.go
sboeuf: Yes I am okay with that. But don't you think we should do that in a following PR as a cleaning purpose ? Because I will have to modify create.go while this PR si supposed to be the exec() implementation ?

sameo: I would appreciate if you could do that as the first commit on this PR.

sboeuf: No problem, will do ;)

create.go
```diff
@@ -81,7 +81,12 @@ func create(containerID, bundlePath, console, pidFilePath string) error {
 	}
 
 	// Start the shim to retrieve its PID.
-	pid, err := startShim(pod)
+	containers := pod.GetContainers()
+	if len(containers) != 1 {
+		return fmt.Errorf("Container list from pod is wrong, expecting only one container")
```
jodh-intel: This indicates a programming error, so I'd prefix with `BUG:` and maybe show how many containers you actually found.

exec.go
```diff
@@ -91,7 +108,118 @@ EXAMPLE:
...
+		params, err := validExecParams(context)
+		if err != nil {
+			return err
+		}
+
+		return execute(params)
 	},
 }
+
+func validExecParams(context *cli.Context) (execParams, error) {
```
jodh-intel: This name is a bit confusing - initially I thought it was going to validate the parameters, but it's actually generating them. Hence, maybe something like `generateExecParams()` would be clearer.

exec.go
```diff
@@ -91,7 +108,118 @@ EXAMPLE:
...
+	}
+
+	// Start the shim to retrieve its PID.
+	pid, err := startShim(process)
+	if err != nil {
+		return err
+	}
+
+	// Creation of PID file has to be the last thing done in the exec
+	// because containerd considers the exec complete after this file
```
jodh-intel: Well, `containerd` considers the `exec` to have "finished starting", not "finished".

shim.go
```diff
@@ -22,57 +22,40 @@ import (
...
+		return -1, fmt.Errorf("Token cannot be empty")
 	}
-	container := containers[0]
 
-	shimInfo, err := getShimInfo(container)
-	if err != nil {
-		return -1, err
+	url := process.URL
+	if url == "" {
+		url = defaultProxyURL
```
jodh-intel: It might be useful to add a debug log call to show which url is actually being used.

create.go
sboeuf: done

exec.go
sboeuf: done

exec.go
sboeuf: done

shim.go
sboeuf: done

shim.go
```diff
@@ -20,59 +20,43 @@ import (
...
+func startShim(process *vc.Process, url string) (int, error) {
+	if process.Token == "" {
+		return -1, fmt.Errorf("Token cannot be empty")
 	}
-	container := containers[0]
 
-	shimInfo, err := getShimInfo(container)
-	if err != nil {
-		return -1, err
+	if url == "" {
```
sameo: This should be an error case. We don't want to support starting a shim without a known URL.

shim.go
```diff
@@ -20,59 +20,43 @@ import (
...
+		url = defaultProxyURL
 	}
+	glog.Infof("Proxy URL: %s\n", url)
 
 	cmd := exec.Cmd{
-		Path: shimInfo.Path,
-		Args: []string{"-t", shimInfo.Token, "-s", shimInfo.IP, "-p", shimInfo.Port},
+		Path: defaultShimPath,
```
sameo: This should be a function argument as well. If our config files specified a shim path that's different than `defaultShimPath` then this is not going to work.
And yes, that means we need to load the config file for `exec` as well.

shim.go
sboeuf: done

shim.go
sboeuf: I agree and all of what you're describing is done here: https://github.com/clearcontainers/runtime/pull/56
Is that okay ?

sameo: @sboeuf: Yes.

sameo:
LGTM

PR #54, opened 2017-03-06
cli: Fix arguments retrieval from CLI

sboeuf:
The way to get arguments from the CLI was not functional. This patch
fixes it for create, delete, kill, start and state actions.

### DIFF
--- a/delete.go
+++ b/delete.go
@@ -43,7 +43,19 @@ EXAMPLE:
 		},
 	},
 	Action: func(context *cli.Context) error {
-		return delete(context.String("container-id"), context.Bool("force"))
+		args := context.Args()
+		if args.Present() == false {
+			return fmt.Errorf("Missing container ID, should at least provide one")
+		}
+
+		force := context.Bool("force")
+		for _, cID := range []string(args) {
+			if err := delete(cID, force); err != nil {
+				return err
+			}
+		}
+
+		return nil
 	},
 }
 
--- a/kill.go
+++ b/kill.go
@@ -43,7 +43,18 @@ EXAMPLE:
 		},
 	},
 	Action: func(context *cli.Context) error {
-		return kill(context.String("container-id"), context.String("signal"))
+		args := context.Args()
+		if args.Present() == false {
+			return fmt.Errorf("Missing container ID")
+		}
+
+		// If signal is provided, it has to be the second argument.
+		signal := args.Get(1)
+		if signal == "" {
+			signal = "SIGTERM"
+		}
+
+		return kill(args.First(), signal, context.Bool("all"))
 	},
 }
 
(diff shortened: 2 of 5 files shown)


delete.go
```diff
@@ -43,7 +43,19 @@ EXAMPLE:
 		},
 	},
 	Action: func(context *cli.Context) error {
-		return delete(context.String("container-id"), context.Bool("force"))
+		args := context.Args()
+		if len(args) < 1 {
+			return fmt.Errorf("No container-id provided, should at least provide one")
+		}
+
```
sameo: This can be replaced by:
```Go
if args.Present() == false {
      return fmt.Errorf("Missing container ID, should at least provide one")
}
```

kill.go
```diff
@@ -43,7 +43,18 @@ EXAMPLE:
 		},
 	},
 	Action: func(context *cli.Context) error {
-		return kill(context.String("container-id"), context.String("signal"))
+		args := context.Args()
+		if len(args) < 1 {
+			return fmt.Errorf("No container-id provided")
+		}
+
```
sameo: Ditto

state.go
```diff
@@ -34,7 +34,12 @@ var stateCommand = cli.Command{
 	Description: `The state command outputs current state information for the
 instance of a container.`,
 	Action: func(context *cli.Context) error {
-		return state(context.String("container-id"))
+		args := context.Args()
+		if len(args) != 1 {
+			return fmt.Errorf("Expecting one argument, got %d arguments: %s", len(args), []string(args))
+		}
+
```
sameo: And this one could be replaced with:

```Go
if args.Tail() !=  nil {
       return fmt.Errorf("Expecting only one container ID, got %d: %v", len(args), []string(args))
}
```

start.go
```diff
@@ -30,7 +32,18 @@ var startCommand = cli.Command{
    unique bon your host.`,
 	Description: `The start command executes the user defined process in a created container .`,
 	Action: func(context *cli.Context) error {
-		return start(context.String("container-id"))
+		args := context.Args()
+		if len(args) < 1 {
+			return fmt.Errorf("No container-id provided, should at least provide one")
+		}
+
```
sameo: Ditto.

start.go
```diff
@@ -30,7 +32,18 @@ var startCommand = cli.Command{
    unique bon your host.`,
```
jodh-intel: An amusing typo here: `bon`! :smile:

sboeuf: I think we should blame @dlespiau for this one ;-)

state.go
sboeuf: This won't work, because args.Tail() will never be nil.

sameo:
LGTM

PR #56, opened 2017-03-08
config: Retrieve shim path from configuration

sboeuf:
We don't want the shim binary path to be hardcoded into the code.
That's why we rely on the configuration file to provide this data.

### DIFF
--- a/config.go
+++ b/config.go
@@ -47,16 +47,22 @@ const (
 	ccProxy = "cc"
 )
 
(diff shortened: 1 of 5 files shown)


sameo:
LGTM

jodh-intel:
lgtm

sameo:
@sboeuf Could you please rebase that one again ?

PR #57, opened 2017-03-08
config: Update proxy config according to virtcontainers

sboeuf:
Proxy config now relies only on a URL provided to the library so
that it can connect to the proxy.


jodh-intel:
lgtm

sameo:
LGTM

PR #58, opened 2017-03-09
Update virtcontainers vendoring

sboeuf:
This PR updates virtcontainers vendoring for following reasons:
- EnterContainer() now returns a Pod pointer.
- Pod has a new public method URL() returning the proxy URL specific to the pod.
- CCProxyConfig contains now only one field: URL.

Then it fixes the build which is broken by this new vendoring.


sameo:
LGTM

PR #60, opened 2017-03-10
cgroups: Create and delete cgroups files related to provided path

sboeuf:
In config.json, a cgroupsPath is provided, and it is expected to be
created. "tasks" and "cgroup.procs" have to be created under this
directory, and filled with the shim's pid. That way, nothing else
will try to create those files on our behalf.

Because we know where we created the files, we are able to remove
them when delete() is called. By removing those files, we make sure
to prevent any file descriptor leak.

### DIFF
--- a/create.go
+++ b/create.go
@@ -18,9 +18,11 @@ package main
 import (
 	"fmt"
 	"os"
+	"path/filepath"
 
 	vc "github.com/containers/virtcontainers"
 	"github.com/containers/virtcontainers/pkg/oci"
+	"github.com/golang/glog"
 	specs "github.com/opencontainers/runtime-spec/specs-go"
 	"github.com/urfave/cli"
 )
@@ -71,7 +73,7 @@ func create(containerID, bundlePath, console, pidFilePath string) error {
 		return err
 	}
 
-	podConfig, shimConfig, _, err := getConfigs(bundlePath, containerID, console)
+	podConfig, shimConfig, ociSpec, err := getConfigs(bundlePath, containerID, console)
 	if err != nil {
 		return err
 	}
@@ -89,6 +91,20 @@ func create(containerID, bundlePath, console, pidFilePath string) error {
 
 	pid, err := startContainerShim(containers[0], shimConfig, pod.URL())
 	if err != nil {
+		return err
+	}
+
+	// config.json provides a cgroups path that has to be used to create "tasks"
+	// and "cgroups.procs" files. Those files have to be filled with a PID, which
+	// is shim's in our case. This is mandatory to make sure there is no one
+	// else (like Docker) trying to create those files on our behalf. We want to
+	// know those files location so that we can remove them when delete is called.
+	cgroupsPath, err := processCgroupsPath(ociSpec)
+	if err != nil {
+		return err
+	}
+
+	if err := createCgroupsFiles(cgroupsPath, pid); err != nil {
 		return err
 	}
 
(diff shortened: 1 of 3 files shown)


create.go
```diff
@@ -115,6 +126,42 @@ func getConfigs(bundlePath, containerID, console string) (vc.PodConfig, ShimConf
...
+		}
+		defer f.Close()
+
+		n, err := f.WriteString(pidStr)
+		if err != nil {
+			return err
+		}
+
+		if n < len(pidStr) {
+			return fmt.Errorf("Could not write pid to '%s': only %d bytes written out of %d", path, n, len(pidStr))
```
jodh-intel: You could use `%q` here (simpler than `'%s'` ;)

create.go
```diff
@@ -91,6 +92,16 @@ func create(containerID, bundlePath, console, pidFilePath string) error {
 		return err
 	}
 
+	// config.json provides a cgroups path that has to be used to create "tasks"
+	// and "cgroups.procs" files. Those files have to be filled with a PID, which
+	// is shim's in our case. This is mandatory to make sure there is no one
+	// else (like Docker) trying to create those files on our behalf. We want to
+	// know those files location so that we can remove them when delete is called.
+	cgroupsPath := "foo"
```
jodh-intel: huh?

create.go
```diff
@@ -115,6 +126,42 @@ func getConfigs(bundlePath, containerID, console string) (vc.PodConfig, ShimConf
 	return *podConfig, shimConfig, nil
 }
 
+func createCgroupsFiles(cgroupsPath string, pid int) error {
```
jodh-intel: This function really needs a test. Something like https://github.com/01org/cc-oci-runtime/blob/master/tests/oci_test.c#L1144..L1198

delete.go
```diff
@@ -94,5 +96,23 @@ func delete(containerID string, force bool) error {
...
+	// function returns.
+	cgroupsPath := "foo"
+	if err := removeCgroupsPath(cgroupsPath); err != nil {
+		return err
+	}
+
 	return nil
 }
+
+func removeCgroupsPath(cgroupsPath string) error {
```
jodh-intel: Again, a brief test function would be good.

oci.go
```diff
@@ -22,6 +22,15 @@ import (
 	vc "github.com/containers/virtcontainers"
 )
 
+// Contants related to cgroup memory directory
+const (
+	cgroupsMemDirPath = "/sys/fs/cgroup/memory"
```
jodh-intel: This might need to become a `var` for the tests.

create.go
sboeuf: Sorry I should have said that the PR is not ready, and that we need to find the way to retrieve cgroupsPath from the config.

create.go
sboeuf: I didn't know that, thanks for the tip ;)

create.go
sboeuf: Agreed.

delete.go
sboeuf: Agreed.

sboeuf:
Be careful to merge #61 before. It is based on it.

delete.go
sboeuf: I am planning to implement a PR about fixing the unit testing in clearcontainers/runtime next week. Is that okay for you if the unit testing of these two functions is done in this next PR ?

create.go
```diff
@@ -92,6 +93,19 @@ func create(containerID, bundlePath, console, pidFilePath string) error {
 		return err
 	}
 
+	// config.json provides a cgroups path that has to be used to create "tasks"
+	// and "cgroups.procs" files. Those files have to be filled with a PID, which
+	// is shim's in our case. This is mandatory to make sure there is no one
+	// else (like Docker) trying to create those files on our behalf. We want to
+	// know those files location so that we can remove them when delete is called.
+	if ociSpec.Linux.CgroupsPath == "" {
+		return fmt.Errorf("CgroupsPath should not be empty")
```
sameo: Are we sure `CgroupsPath` is an OCI MUST ? Surely Docker uses it, but I don't see where it's stated as a mandatory setting in OCI.
@jodh-intel Looking for your OCI expert input here...

jodh-intel: @sameo - good catch. You're right - if it's specified, the runtime needs to honor it, but it's not mandatory:

- https://github.com/opencontainers/runtime-spec/blob/master/config-linux.md#control-groups

create.go
```diff
@@ -116,6 +130,42 @@ func getConfigs(bundlePath, containerID, console string) (vc.PodConfig, ShimConf
 	return *podConfig, shimConfig, *ociSpec, nil
 }
 
+func createCgroupsFiles(cgroupsPath string, pid int) error {
+	if cgroupsPath == "" {
+		return fmt.Errorf("Missing cgroups path")
+	}
+
+	cgroupsDir := filepath.Join(cgroupsMemDirPath, cgroupsPath)
```
jodh-intel: ... and looking at the spec on cgroups again more closely, we should check whether `cgroupsPath` is absolute or relative. If we don't care, we should still document that approach.

create.go
sboeuf: @sameo yes you're right, it is not mandatory, there is no reason for returning an error.
@jodh-intel I will follow the spec on this one, I will try to handle both relative and absolute paths ;)

create.go
grahamwhaley: and make sure we understand if there is a difference between 'the spec' and what 'runc' does/supports. :-)

sameo:
LGTM. Unless @jodh-intel wants tests right away, it's ready to be merged.

jodh-intel:
Hi @sameo, @sboeuf  - I'm happy to wait for a follow-on PR.

lgtm.

PR #65, opened 2017-03-14
shim: Fix execution of cc-shim binary

sboeuf:
The use of exec.cmd was not proper, causing the binary to be
re-executed. Relying on the exec.Command() function allows to
start cc-shim without re-executing it.

### DIFF
--- a/shim.go
+++ b/shim.go
@@ -42,11 +42,8 @@ func startShim(process *vc.Process, config ShimConfig, url string) (int, error) 
 	}
 	glog.Infof("Shim binary path: %s\n", config.Path)
 
-	cmd := exec.Cmd{
-		Path: config.Path,
-		Args: []string{"-t", process.Token, "-u", url},
-		Env:  os.Environ(),
-	}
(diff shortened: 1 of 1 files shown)


sboeuf:
@jodh-intel @sameo @dlespiau please take a look.

jodh-intel:
Out of interest, how did you find this issue? `exec.Cmd` is also used by `virtcontainers`.

lgtm

dlespiau:
I would wager the unit tests in the runtime :)

sboeuf:
@jodh-intel I found it by using the new mock binary I have created for shim's unit tests. In virtcontainers, that is different. We want to re-exec the binary in case of CNM network, that is we need to re-exec dockerd.

jodh-intel:
@dlespiau - in which case, we'd have seen travis failures right? I don't see any...?

jodh-intel:
@sboeuf - ah! thanks for the info!

sboeuf:
@jodh-intel no wee didn't see any error from unit tests because we only start the shim binary, we don't check that it is not failing after it started running.

PR #66, opened 2017-03-17
ci: Display unified diff with what gofmt wants us to change

dlespiau:
We run gofmt with the simplify (-s) option. Some of the rewrite rules are
hidden behind that flag because they break previous go versions. Editors
don't run gofmt with -s by default for that reason and so it's not
unlikely the static checks scripts find something to change.

We can do better than displaying the list of "offending" files, we can
make gofmt display the unified diff with the suggested simplification.
It's just a matter of using the -d flag.

Signed-off-by: Damien Lespiau <damien.lespiau@intel.com>


jodh-intel:
lgtm (again ;)

PR #67, opened 2017-03-31
Fix runtime due to shim/proxy 3.0 enabling

sboeuf:
Here we have a list of patches aiming to fix the runtime so that it can properly work with shim/proxy 3.0.

### DIFF
--- a/config.go
+++ b/config.go
@@ -55,18 +55,24 @@ const (
 	ccShim = "cc"
 )
 
+const (
+	hyperstartAgent = "hyperstart"
+)
+
 var (
 	errUnknownHypervisor  = errors.New("unknown hypervisor")
 	errUnknownAgent       = errors.New("unknown agent")
 	errTooManyHypervisors = errors.New("too many hypervisor sections")
 	errTooManyProxies     = errors.New("too many proxy sections")
 	errTooManyShims       = errors.New("too many shim sections")
+	errTooManyAgents      = errors.New("too many agent sections")
 )
 
 type tomlConfig struct {
 	Hypervisor map[string]hypervisor
 	Proxy      map[string]proxy
 	Shim       map[string]shim
+	Agent      map[string]agent
 }
 
 type hypervisor struct {
@@ -81,6 +87,10 @@ type proxy struct {
 
 type shim struct {
 	Path string `toml:"path"`
+}
+
+type agent struct {
+	PauseRootPath string `toml:"pause_root_path"`
 }
 
 func (h hypervisor) path() string {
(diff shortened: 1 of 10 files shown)


oci.go
```diff
@@ -129,16 +129,66 @@ func stopContainer(podStatus vc.PodStatus) error {
...
+
+		cgroupsPathList = append(cgroupsPathList, blkIOCgroupsPath)
+	}
+
+	return cgroupsPathList, nil
+}
+
+func processCgroupsPathForResource(ociSpec specs.Spec, resource string) (string, error) {
+	if resource == "" {
+		return "", fmt.Errorf("resource is empty")
```
dlespiau: All these could be replaced by errors.New(), that avoids Errorf scanning the format string looking of % markers, not that we really care about speed in error paths...

oci.go
```diff
@@ -129,16 +129,66 @@ func stopContainer(podStatus vc.PodStatus) error {
...
 	if ociSpec.Linux.CgroupsPath == "" {
-		return "", nil
+		return []string{}, nil
+	}
+
+	if ociSpec.Linux.Resources == nil {
+		return []string{}, nil
+	}
+
+	if ociSpec.Linux.Resources.Memory != nil {
```
dlespiau: maybe `for resource := range [...]string{"memory", "cpu", ...}`

oci.go
sboeuf: Yes I can change this.

oci.go
sboeuf: Well I cannot do that because for instance in case of memory, there is nothing to help making the relation between the string "memory" and the structure Memory. Am I missing something ?

dlespiau: ah, fair enough I missed this. it's always possible to abstract that but probably not worth the effort.

config.go
```diff
@@ -35,6 +35,7 @@ const (
 	defaultImagePath            = "/usr/share/clear-containers/clear-containers.img"
 	defaultHypervisorPath       = "/usr/bin/qemu-lite-system-x86_64"
 	defaultProxyURL             = "unix:///run/cc-oci-runtime/proxy.sock"
+	defaultPauseBinPath         = "/tmp/bundles/pause_bundle/rootfs/bin/pause"
```
jodh-intel: Really? Why does it have to be here?

sboeuf: Well this file is supposed to read the configuration for our runtime. This is part of the configuration to indicate where we can find the pause binary, isn't it ?

jodh-intel: But why is the default below `/tmp/`?

sboeuf: Yes I agree it would be better somewhere else. But the goal of this PR was to get everything working. What are you suggesting as a default location ?
And as a follow up PR, we could have this path as a parameter in the toml config file.

jodh-intel: A path like `/var/lib/clearcontainers/runtime/bundles/pause_bundle/rootfs/bin/pause` would probably be more inline with the FHS (although I don't like that it hard-codes the bundle name).

Going forward, I can't see why -- as you say -- we can't just read the bundle path from the config file. The pause bin path would then be relative to that. The current relative path (`/bin/pause`) should probably be changed to make it "more unique" as it's currently a pretty generic name. Something like `/usr/libexec/clearcontainers/runtime/pause-container` for example would be more descriptive and much less likely to conflict with existing bundle contents.

Finally, shouldn't there be a check on whether the bundle directory part of `defaultPauseBinPath` exists, since if it doesn't it's not (ever?) going to be valid.

sboeuf: @jodh-intel I can do this change, I wanted to do that anyway.
@sameo should I add a default binary pause path or we can consider we have to read it from the config file ? In case you want a default one, what do you think as a default location ?

sameo: @sboeuf it would be better to add the pause bundle path as an optional runtime config and have a sane default for it (`/var/lib/clearcontainers/runtime/bundles/pause_bundle` looks ok).

Then the pause binary path should be relative to that and although I agree with @jodh-intel about it being quite generic, the pause bundle is explicitly made of one single binary file so the risks of seeing a conflict with the rest of the bundle content is really low. Given that `bin/pause` is a k8s accepted standard, I think we're pretty safe keeping it as is.

sboeuf: @sameo Just to make sure I understand correctly. We want `/var/lib/clearcontainers/runtime/bundles/pause_bundle` as default, and for example, the config.toml file will provide `rootfs/bin/pause` as a relative path to add to this default one ?
You mentioned this as optional, but it has to be mandatory because in case of runtime implementation, we will use hyperstart, meaning we must provide a pause binary path for the pause container. Do you agree ?
As a global comment, I don't understand why we are not providing the full path here through the config.toml file. Since we have to know the full path, there is no point in having one part of the path hardcoded by default and another part provided by the config file.

sameo: @sboeuf sorry if my answer was confusing but I'd like:

1. The pause **bundle** path to be optionally defined from `config.toml` and use the default `/var/lib/clearcontainers/runtime/bundles/pause_bundle` if not specified. In any case it must be an aboslute path, obviously.
2. Use `bin/pause` as the non configurable relative path to the pause binary.

Does that make more sense ?

sboeuf: Yes it makes more sense in that case :)

sboeuf:
@sameo @jodh-intel @dlespiau All comments addressed, let me know if you're okay with this new version ;)

config.go
```diff
@@ -192,19 +188,95 @@ func loadConfiguration(configPath string) (oci.RuntimeConfig, ShimConfig, error)
...
+
+			break
+		}
+	}
+
+	for k, agent := range tomlConf.Agent {
+		switch k {
+		case hyperstartAgent:
+			path := filepath.Join(agent.pauseBundlePath(),
+				pauseBinRelativePath)
```
jodh-intel: Isn't this missing the `/rootfs/` directory?

config.go
```diff
@@ -35,6 +36,8 @@ const (
 	defaultImagePath            = "/usr/share/clear-containers/clear-containers.img"
 	defaultHypervisorPath       = "/usr/bin/qemu-lite-system-x86_64"
 	defaultProxyURL             = "unix:///run/cc-oci-runtime/proxy.sock"
+	defaultPauseBinBundlePath   = "/var/lib/clearcontainers/runtime/bundles/pause_bundle"
+	pauseBinRelativePath        = "bin/pause"
```
jodh-intel: This looks good.

config.go
sboeuf: Well I didn't want to add this "rootfs" inside the code, I think it is better to have it from the absolute bundle path we are providing from configuration.toml

jodh-intel: I agree it's best not to hard-code, but an OCI bundle contains a `config.json` and a rootfs directory. Conventionally the latter is called `rootfs/`, but it doesn't have to be. Strictly you need to check the `root.path` element in `config.json` to know the actual relative path to the bundles rootfs directory. But the code now seems to just glue the bundle path and the pause bin path together which looks wrong (previously you did have a `rootfs/` directory element).

sboeuf: But this is specific to an OCI bundle, thus I don't see the point in following those conventions. And this rootfs value won't come from a config.json because it won't come from Docker for instance. This is specific to our use of hyperstart. What do you suggest to retrieve this value from somewhere ?
@sameo any suggestion on this one ?

sameo: @sboeuf @jodh-intel this is hyperstart specific and we only need a statically linked pause binary to give to hyperstart so that it can create the pause container in the pod.
Thus an OCI config file is not needed and we can just glue the bundle path (bundle may be the wrong term here...) with the pause binary relative path.

jodh-intel: @sameo - ah - thanks for the clarification. In which case I guess this is fine ;)

sboeuf:
@sameo @jodh-intel new version of the PR, replacing "bundle" with "root" as it is more a pause root path than a pause bundle path. Let me know if this works for you and feel free to merge !

sameo:
LGTM

jodh-intel:
lgtm

PR #68, opened 2017-04-06
Fix runtime and improve logger

sboeuf:
This PR first fixes a mistake coming from a previous one. Then it makes the runtime more consistent by using logrus everywhere, instead of glog.


amshinde:
lgtm

PR #71, opened 2017-04-14

sboeuf:
Virtcontainers recently changed to include 2 important fixes:
- Proxy does not need a token for pause container
- Add LockOSThread() in an init() function, to ensure the main thread is not executed in a thread ID different from the process ID.


dlespiau:
lgtm

PR #72, opened 2017-04-14

sboeuf:
Include one important fix:
- cc_proxy implementation was marshalling the message sent to the proxy while the proxy was also marshalling it. This error was not detected before because Go 1.8 was working fine with a double
marshalling.


sameo:
LGTM

PR #73, opened 2017-04-19
Improve logging

jodh-intel:


jodh-intel:
An alternative to writing to syslog is to write to a global log as we do for https://github.com/01org/cc-oci-runtime.

The complication with that approach though is the need to pass the `--global-log=` option to that runtime (see https://github.com/01org/cc-oci-runtime/wiki/Debugging).

dlespiau:
I'm a bit concerned that we'll end up spamming syslog with many messages that are really debug messages. What about hiding with an env variable and/or command line parameter?

jodh-intel:
I was thinking something similar last night. As mentioned, my thoughts were to not add a new CLI param as that would involve a helper script (like `cc-oci-runtime.sh`) which isn't ideal.

I'm now thinking I could add a `global_log` entry under a new `[runtime]` section in `configuration.toml`. But also honour an env-var as you suggest.

dlespiau:
ah yes! a config option sounds like a very good way to do it.

jodh-intel:
Closing. 

I've split two of these commits into #78 and #79 and will create a PR for a global log option instead.

jodh-intel:
Global logger in now #80.

PR #77, opened 2017-04-20
Add standard errors and arg checks

jodh-intel:


dlespiau:
I'm not a big fan of those :(

`config` being nil is a programming error, so letting it panic is the right thing to do as far as I'm concerned. Another way to look at it is that those code paths aren't protecting an input of the program but an are an internal consistency check (where we'd use an assert in C).

Validating input config values is definitely a good thing though.

jodh-intel:
So, drop the checks on pointers being `nil`?

dlespiau:
I guess, yes, it means dropping the nil checks.

The Token comes from the proxy, so checking that input makes sense to me (but only where we effectively get the data for the first time).

A nice note is that there's nothing like malloc() able to return NULL in go (that's not entirely true very large allocation attempts can panic, which can be recovered).

It's quite a change from 2.1, happy to get more opinions on this :)

jodh-intel:
Branch updated.

jodh-intel:
@sboeuf - any comments on this?

create.go
```diff
@@ -72,6 +72,10 @@ func create(containerID, bundlePath, console, pidFilePath string) error {
 		return err
 	}
 
+	if pidFilePath == "" {
```
dlespiau: This one isn't necessary. We check it already in 

```
func createPIDFile(pidFilePath string, pid int) error {
	if pidFilePath == "" {
		return errNeedPidFilePath
	}
```

I would also argue that second one totally isn't necessary either. One of the functions using pidFilePAth will return an error.

create.go
```diff
@@ -115,6 +119,14 @@ func create(containerID, bundlePath, console, pidFilePath string) error {
 }
 
 func getConfigs(bundlePath, containerID, console string) (vc.PodConfig, specs.Spec, error) {
+	if bundlePath == "" {
```
dlespiau: Not sure about checking those values again in an internal functions when we have things like validCreateParams() that seems like the place to have those checks.

kill.go
```diff
@@ -102,6 +102,10 @@ func kill(containerID, signal string, all bool) error {
 		return err
 	}
 
+	if signal == "" {
```
dlespiau: Here again we do the checking outside of this function already, ensuring signal cannot be ""

oci.go
```diff
@@ -34,13 +33,13 @@ const (
...
 )
 
-var (
-	errNeedLinuxResource = errors.New("Linux resource cannot be empty")
-)
-
 var cgroupsMemDirPath = "/sys/fs/cgroup"
 
 func containerExists(containerID string) (bool, error) {
+	if containerID == "" {
```
dlespiau: I can't see the value of having a special case for "". containerExists() will already return false. Somethig that would add value though is returning an error when returning false in this function, right now we `return false, nil` and that's clearly a bug.

oci.go
```diff
@@ -58,7 +57,12 @@ func containerExists(containerID string) (bool, error) {
 func validCreateParams(containerID, bundlePath string) error {
 	// container ID MUST be provided.
 	if containerID == "" {
-		return fmt.Errorf("Missing container ID")
+		return errNeedContainerID
```
dlespiau: Here again, containerExists should fail with a nice error message including the containerId instead of special casing the "" value.

oci.go
dlespiau: Argh. Got that wrong.

That said, we're not checking at several places that containerID is not "". One, at the start, should be enough, not at the start + in containerExists

create.go
jodh-intel: I appreciate that checks are performed later on, but the point is that *work has been done **before** those checks are carried out*. Work that we only find out should **not** have been done much later on in the program. Are you suggesting that this sort of approach is "the golang way"?

dlespiau: I try not to do generalities like "the go way" or "unix way", that tends to be more dogmatic than useful :)

What I'm trying to say: check constraints (can be also worded as establish strong invariants) as early as possible: this could be either program inputs are library entry points. Then, in the rest of the code, these invariants are established and don't need to be checked again.

It sometimes useful to have the nil value (0, nil, "") a valid input, that's a different thing though. That will lead to internal checks for nil, but that's just because it's a valid input and we need to do something special for that case.

sboeuf:
@jodh-intel @dlespiau Honestly it is up to you guys on this one. I agree that we don't need to "overcheck" all inputs all the time, but as long as it is not hurting the program. And I think it is pretty low impact for performances.
As long as @dlespiau comments are addressed, this LGTM.

sboeuf:
@jodh-intel Ping on this one ! What is the status on this one ?

jodh-intel:
@sboeuf - I'm honestly not sure what's going on with it. I get the impression there is no perceived value in these checks (?)

dlespiau:
Still not convinced by the systematic approach. It looks like the equivalent of sprinkling `if (p == NULL) return;` in C to me. Especially the cases where we end up testing the same string for "" multiple times in the same code path.

sboeuf:
@jodh-intel well I like the idea of having errors.go so that we factorize error messages but we have to agree where the checks have to be performed. This should happen at the higher level possible for each variable. This means that you don't want to test containerID after the check has been performed by validCreateParams() (in case of create for instance).
I don't think there is a generic pattern here, we just need to understand the code and make sure variables are valid asap.
How does that sound ?

sboeuf:
@jodh-intel I guess you should drop this one. It's been more than a month since you submitted it.

jodh-intel:
agreed.

PR #80, opened 2017-04-20
Add global logfile

jodh-intel:


jodh-intel:
@sboeuf - would you mind reviewing please?

oci.go
```diff
@@ -41,6 +41,11 @@ var (
 var cgroupsMemDirPath = "/sys/fs/cgroup"
 
 func containerExists(containerID string) (bool, error) {
+
```
dlespiau: As said elsewhere, the real bug is in the final return:

```
return false, nil
```

dlespiau: Ah no, I got that totally wrong. The behaviour of this function is, huuuum, interesting :)

sboeuf: @dlespiau What is not clear in that function ??

jodh-intel: I think this summarises to a nop for this PR? Maybe @dlespiau can raise an issue/PR for any problems with `containerExists()`?

dlespiau: Now that I look at it again, nothing seems weird. For some reason I didn't expect to have `false, nil` a valid combination, brain fart I guess.

sboeuf:
Okay I did the review there is one thing I would like to discuss. You have to call loadConfiguration() from the main.go now, meaning it is called all the time, for every command.
First, I have realized that given the recent changes related to the shim, we don't need to call loadConfiguration() from exec. This means we only need it into create.go. Now, knowing that, don't you think it will impact performances to parse/load this configuration for every command while we need it only for one ?

@sameo @dlespiau please give us your opinion on this !

If we are fine with this, then this PR LGTM.

jodh-intel:
Hi @sboeuf - I agree in principle that it might seem wasteful, yes. However, I'd say it was very unusual for a binary to only parse its configuration for specified sub-commands: it's much more usual to have `main()` load the config file first. Added to which, we may well want to add further options to the config file to avoid having to use a "helper script" as we currently do for `cc-oci-runtime` (see https://github.com/01org/cc-oci-runtime/wiki/Debugging). If we find that parsing a small toml file is too slow, I'd suggest we find/write a faster parser.

We **could** rely entirely on the environment variable, but will a major restriction: it won't work under docker since even if the user remembers to add `-e` to their `docker run` (`docker run -e CC_RUNTIME_GLOBAL_LOG=/tmp/runtime.log`), the runtime won't have that variable set. So even if we modify the runtime to check for that variable in `config.json` (yuck ;-) and have it enable debugging at that point, we cannot have debug enabled from runtime startup which restricts the feature significantly.

dlespiau:
I think the only way to go with performance is benchmarking :) parsing a file compared to booting a VM can't be much work. It's probably less work than the dynamic loader is doing to execute the runtime.

When we have good benchmarks showing where we spend time, we can improve on many subjects, included this one if needed.

jodh-intel:
+1, but I think adding this generally useful feature should happen sooner rather than later to make debugging actually possible under docker ;-)

When the runtime is fully usable we can start to optimise based on perf benchmarks.

sboeuf:
Okay I think we still have a lot of development to do, meaning this debug feature will be nice, and we can still end up removing it eventually.

sboeuf:
LGTM

PR #86, opened 2017-04-24
exec: Unmarshal to a specs.Process pointer

sboeuf:
The unmarshalling was failing because we were not passing a pointer to a specs.Process structure.
With this patch, the exec command works as expected.


sameo:
LGTM

PR #89, opened 2017-04-29
Get the runtime tested with Semaphore CI

sboeuf:
This PR allows to test the entire CC 3.0 chain with Docker commands. This introduces Semaphore CI which supports nested VMs.


.ci/install_cc_env_ubuntu.sh
```diff
@@ -0,0 +1,69 @@
...
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+
+#!/bin/bash
+
+function check_err {
```
jodh-intel: You can drop this function if you add `set -e` at the top of the script`.

.ci/install_cc_env_ubuntu.sh
```diff
@@ -0,0 +1,69 @@
...
+		exit 1
+	fi
+}
+
+echo "Add clear containers sources to apt list"
+sudo sh -c "echo 'deb http://download.opensuse.org/repositories/home:/clearlinux:/preview:/clear-containers-2.1/xUbuntu_16.10/ /' >> /etc/apt/sources.list.d/cc-oci-runtime.list"
+check_err
+
+echo "Update apt repositories"
+sudo apt-get update
```
jodh-intel: Not essential but you might want to add `-qq` to this call to discard output. Alternatively, you could use `chronic` as we do here:

https://github.com/01org/cc-oci-runtime/blob/master/.ci/ci-configure.sh#L44..L46

.ci/install_cc_env_ubuntu.sh
```diff
@@ -0,0 +1,69 @@
...
+check_err
+
+echo "Install systemd"
+sudo apt-get install -y systemd-services
+check_err
+
+echo "Install linux-container kernel"
+sudo apt-get install -y --force-yes linux-container
```
jodh-intel: OOI, do we actually need `--force-yes` here? You already specify `-y` so that should be sufficient shouldn't it?

.ci/install_cc_env_ubuntu.sh
```diff
@@ -0,0 +1,69 @@
...
+sudo apt-get install -y --force-yes qemu-lite
+check_err
+
+echo "Download clear containers image"
+latest_version=$(curl -sL https://download.clearlinux.org/latest)
+curl -LO "https://download.clearlinux.org/current/clear-${latest_version}-containers.img.xz"
+check_err
+
+echo "Extract clear containers image"
+unxz clear-${latest_version}-containers.img.xz
```
jodh-intel: Missing checksum test. You can copy that from here:

https://github.com/01org/cc-oci-runtime/blob/master/.ci/ci-setup.sh#L184..L209

.ci/install_cc_proxy.sh
```diff
@@ -0,0 +1,37 @@
...
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+
+#!/bin/bash
+
+function check_err {
```
jodh-intel: `set -e`.

.ci/install_cc_runtime.sh
```diff
@@ -0,0 +1,78 @@
...
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+
+#!/bin/bash
+
+function check_err {
```
jodh-intel: `set -e`.

.ci/install_cc_runtime.sh
```diff
@@ -0,0 +1,78 @@
...
+check_err
+
+echo "Restart docker service"
+sudo service docker stop
+sudo service docker start
+check_err
+
+echo "Install cc-runtime configuration.toml to /etc/clear-containers"
+sudo mkdir -p /etc/clear-containers
+sudo cp config/configuration_w_logs.toml /etc/clear-containers/configuration.toml
```
jodh-intel: Does that `w` mean "with"? If so, I'd rename it `configuration-with-logs.toml` :smile:

.ci/install_cc_runtime.sh
```diff
@@ -0,0 +1,78 @@
...
+make
+check_err
+
+echo "Install cc-runtime"
+sudo make install
+check_err
+
+echo "Add cc-runtime as new Docker runtime"
+sudo mkdir -p /etc/default
+cat << EOF | sudo tee /etc/default/docker
```
jodh-intel: Which version of docker are you testing with?

.ci/install_cc_env_ubuntu.sh
```diff
@@ -0,0 +1,69 @@
...
+check_err
+
+sudo mkdir -p /usr/share/clear-containers/
+echo "Install clear containers image"
+sudo install --owner root --group root --mode 0755 clear-${latest_version}-containers.img /usr/share/clear-containers/
+check_err
+
+echo "Create symbolic link /usr/share/clear-containers/clear-containers.img"
```
jodh-intel: You've specified it twice, so I'd suggest using a variable for this path to avoid hard-coding `/usr/share/clear-containers/clear-containers.img` in the string.

.ci/install_cc_env_ubuntu.sh
```diff
@@ -0,0 +1,69 @@
...
+curl -LO "https://download.clearlinux.org/current/clear-${latest_version}-containers.img.xz"
+check_err
+
+echo "Extract clear containers image"
+unxz clear-${latest_version}-containers.img.xz
+check_err
+
+sudo mkdir -p /usr/share/clear-containers/
+echo "Install clear containers image"
+sudo install --owner root --group root --mode 0755 clear-${latest_version}-containers.img /usr/share/clear-containers/
```
jodh-intel: Maybe create a variable for `/usr/share/clear-containers/` to avoid duplication and make maintenance easier?

.ci/install_cc_env_ubuntu.sh
```diff
@@ -0,0 +1,69 @@
...
+check_err
+
+echo "Create symbolic link /usr/share/clear-containers/clear-containers.img"
+sudo ln -fs /usr/share/clear-containers/clear-${latest_version}-containers.img /usr/share/clear-containers/clear-containers.img
+check_err
+
+echo "Install clear containers kernel 4.5-50"
+sudo install --owner root --group root --mode 0755 .ci/kernel/vmlinux-4.5-50.container /usr/share/clear-containers/
```
jodh-intel: Why are you bundling a binary kernel version? I really don't think this is a good idea. I'd recommend downloading the pre-built latest kernel from OBS. There are no packages for Trusty, but the `.deb` files for the kernel should be distro-version-agnostic.

.ci/install_cc_env_ubuntu.sh
sboeuf: What if I want to use this script separately ?I think it is better to keep this function inside every script so that we can use it separately if needed.

.ci/install_cc_proxy.sh
sboeuf: Ditto

.ci/install_cc_runtime.sh
sboeuf: Ditto.

.ci/install_cc_runtime.sh
sboeuf: It is 1.12.6, which is the default one with Semaphore CI. I think a second step would be to try downloading and replacing this default version, but as a first step I am relying on the default version already provided by Semaphore.

.ci/install_cc_runtime.sh
sboeuf: Yes I will do that ;)

.ci/install_cc_env_ubuntu.sh
sboeuf: Makes sense, will do.

.ci/install_cc_env_ubuntu.sh
jodh-intel: I don't really understand - if you want to run this script separately you can but you are re-implementing a feature that is already in the shell. If you wanted to, you could add a command-line option/argument that conditionally called `set -e` I guess?

.ci/install_cc_env_ubuntu.sh
sboeuf: Well for some reasons that I don't know/understand, we have some kernel panic inside the VM when using an clear containers kernel >= 4.9. The most recent before 4.9 being 4.5-50, that's why I took this kernel. We already discussed this with @sameo as a temporary solution, because we don't really care about the kernel we are using for CI right now, as long as the CI run properly.
But this needs to be investigated to find out the reason of the crash.

.ci/install_cc_runtime.sh
jodh-intel: It's worth pointing out that that default might change "underneath you". That's why we have https://github.com/01org/cc-oci-runtime/blob/master/versions.txt#L11 to ensure we know what we're testing with.

.ci/install_cc_env_ubuntu.sh
sboeuf: It is about untrusted sources, that's the way we don't get prompted about that.

.ci/install_cc_env_ubuntu.sh
sboeuf: Thanks.

.ci/install_cc_env_ubuntu.sh
dlespiau: (probably a good idea to file a bug about it)

jodh-intel: @sboeuf - this sort of change tends to get forgotten about very quickly. I'd still suggest downloading the kernel at the version you need, but in this case, you probably want a big fat multi-line warning message displayed to ensure it is clear when reviewing logs that a fixed and old version of the kernel is being used. That message should display the bug number @dlespiau's suggests you raise. Something like:

```
echo -e "\nWARNING:"
echo "WARNING: Using backlevel kernel version '$kernel_version' due to bug $bug_url"
echo -e "WARNING:\n"
```

sboeuf: Yes I can do that, it will be a bit cleaner. https://github.com/clearcontainers/runtime/issues/91 is the issue I have just opened.

.ci/install_cc_runtime.sh
sboeuf: I can add a comment on this one but I think we really need a follow up PR to be able to build and test Docker version we want.

sboeuf:
@jodh-intel I think I have addressed all your comments. Let me know if I forgot something !

oci.go
```diff
@@ -32,6 +32,7 @@ const (
 	cgroupsDirMode   = os.FileMode(0750)
 	cgroupsFileMode  = os.FileMode(0640)
 	cgroupsMountType = "cgroup"
+	cgroupFsType     = 0x27e0eb
```
devimc: what filesystem is this?

sboeuf: cgroup. Look here http://man7.org/linux/man-pages/man2/statfs.2.html (CGROUP_SUPER_MAGIC)

grahamwhaley: Feels like if there is no go lib that proves that constant, then at least a comment to explain the magic is in order :-)

sboeuf:
@grahamwhaley I addressed your comment ;)

Makefile
```diff
@@ -18,5 +22,8 @@ check-go-test:
 check-go-static:
 	.ci/go-static-checks.sh $(GO_STATIC_CHECKS_ARGS)
 
+install:
+	$(QUIET_INST)install -D $(TARGET) $(DESTDIR)$(LIBEXECDIR)/clearcontainers/$(TARGET) || exit 1;
```
amshinde: Why is this being installed in libexec dir? The runtime can be called standalone, so it would be make sense to install this under /usr/bin so that it is found in the search path. Also, this makes it possible to add the runtime to docker without specifying the complete path.

.ci/install_virtcontainers.sh
```diff
@@ -0,0 +1,33 @@
...
+set -e
+
+echo "Retrieve virtcontainers repository"
+go get -d github.com/containers/virtcontainers
+pushd $GOPATH/src/github.com/containers/virtcontainers
+
+echo "Build pause binary"
+make pause
+
+echo "Create /var/lib/clearcontainers/runtime/pause_bundle/bin"
```
amshinde: The path being created and the echo message are not the same. The message has missing "bundles"

sboeuf: Thanks for catching this.

Makefile
sboeuf: Yes that makes sense. Let's use /usr/local/bin instead.

sboeuf:
@amshinde I have addressed your comments.

sboeuf:
@jodh-intel if you're okay with the latest version of that PR, please merge it ;)

.ci/install_cc_runtime.sh
```diff
@@ -0,0 +1,61 @@
...
+DOCKER_OPTS="-D --add-runtime cor=/usr/local/bin/cc-runtime --default-runtime=cor"
+EOF
+
+echo "Restart docker service"
+sudo service docker stop
+sudo service docker start
+
+echo "Install cc-runtime configuration.toml to /etc/clear-containers"
+sudo mkdir -p /etc/clear-containers
+sudo cp config/configuration_with_logs.toml /etc/clear-containers/configuration.toml
```
jodh-intel: Thanks for renaming this config file. But on reflection, I think it would be safer to only have a single config file and modify it as needed. That will make maintenance easier and minimise the chance of "code rot" if we add new options to the main config file but forget to update the "special" one for testing).

For example, I do this:

```
sed 's/^#\(\[runtime\]\|global_log_path =\)/\1/g' config/configuration.toml | sudo tee /etc/clear-containers/configuration.toml
```

.ci/install_virtcontainers.sh
```diff
@@ -0,0 +1,33 @@
...
+pushd $GOPATH/src/github.com/containers/virtcontainers
+
+echo "Build pause binary"
+make pause
+
+echo "Create /var/lib/clearcontainers/runtime/bundles/pause_bundle/bin"
+sudo mkdir -p /var/lib/clearcontainers/runtime/bundles/pause_bundle/bin
+
+echo "Install pause binary"
+sudo cp pause/pause /var/lib/clearcontainers/runtime/bundles/pause_bundle/bin/
```
jodh-intel: This should really be `sudo install --owner ... --group ... --mode ... pause/pause ...`

.ci/install_virtcontainers.sh
```diff
@@ -0,0 +1,33 @@
...
+
+echo "Retrieve virtcontainers repository"
+go get -d github.com/containers/virtcontainers
+pushd $GOPATH/src/github.com/containers/virtcontainers
+
+echo "Build pause binary"
+make pause
+
+echo "Create /var/lib/clearcontainers/runtime/bundles/pause_bundle/bin"
+sudo mkdir -p /var/lib/clearcontainers/runtime/bundles/pause_bundle/bin
```
jodh-intel: Another opportunity to use a variable for these repeated paths.

sboeuf:
@jodh-intel Comments addressed !

jodh-intel:
lgtm

PR #90, opened 2017-05-02
Makefile: Add an install target for runtime binary

amshinde:
We were missing an install target, add this.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,3 +1,5 @@
+PREFIX := $(DESTDIR)/usr
+BINDIR := $(PREFIX)/bin
 SOURCES := $(shell find . 2>&1 | grep -E '.*\.(c|h|go)$$')
 VERSION := ${shell cat ./VERSION}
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -1,3 +1,5 @@
+PREFIX := $(DESTDIR)/usr
```
dlespiau: DESTDIR isn't part of the prefix, it's part of the install target. ie if we use $(PREFIX)/share/clear-containers as a string in the binary, we don't want DESTDIR in there.

```
PREFIX=/usr
...
install:
        install -D $(TARGET) $(DESTDIR)/$(BINDIR)/$(TARGET)
```

sboeuf:
@amshinde @dlespiau This patch is already part of this PR https://github.com/clearcontainers/runtime/pull/89

amshinde:
@sboeuf  I'll close this then.

PR #92, opened 2017-05-03
VERSION: Move back to alpha

sameo:
We're much closer to an alpha than to a release candidate.

Signed-off-by: Samuel Ortiz <sameo@linux.intel.com>


jodh-intel:
lgtm

PR #107, opened 2017-05-03
kill: adds support for numeric signals

devimc:
With this commit kill command is able to handle numeric
signals as argument

### DIFF
--- a/kill.go
+++ b/kill.go
@@ -17,6 +17,7 @@ package main
 
 import (
 	"fmt"
+	"strconv"
 	"syscall"
 
 	vc "github.com/containers/virtcontainers"
@@ -129,9 +130,44 @@ func kill(containerID, signal string, all bool) error {
 		return fmt.Errorf("Process not running inside container %s", containerID)
 	}
 
-	if err := vc.KillContainer(containerID, podStatus.ContainersStatus[0].ID, signals[signal]); err != nil {
+	signum, err := processSignal(signal)
+	if err != nil {
+		return err
+	}
+
+	if err := vc.KillContainer(containerID, podStatus.ContainersStatus[0].ID, signum); err != nil {
 		return err
 	}
 
 	return nil
+}
+
+func processSignal(signal string) (syscall.Signal, error) {
+	signum, signalOk := signals[signal]
+	if signalOk {
+		return signum, nil
+	}
+
+	// Support for short name signals (INT)
+	signum, signalOk = signals["SIG"+signal]
+	if signalOk {
+		return signum, nil
+	}
+
+	// Support for numeric signals
+	s, err := strconv.Atoi(signal)
+	if err != nil {
+		return 0, fmt.Errorf("Failed to convert signal %s to int", signal)
+	}
+
+	signum = syscall.Signal(s)
+	// Check whether signal is valid or not
+	for _, sig := range signals {
+		if sig == signum {
+			// signal is a valid signal
+			return signum, nil
+		}
+	}
+
+	return 0, fmt.Errorf("Signal %s is not supported", signal)
 }
--- a/kill_test.go
+++ b/kill_test.go
@@ -1,0 +1,48 @@
+// Copyright (c) 2017 Intel Corporation
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//      http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package main
+
+import (
+	"syscall"
+	"testing"
+)
+
+func TestProcessSignal(t *testing.T) {
+	tests := []struct {
+		signal string
+		valid  bool
+		signum syscall.Signal
+	}{
+		{"SIGDCKBY", false, 0}, //invalid signal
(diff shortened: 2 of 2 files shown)


kill.go
```diff
@@ -129,7 +130,27 @@ func kill(containerID, signal string, all bool) error {
...
+		s, err := strconv.Atoi(signal)
+		if err != nil {
+			return fmt.Errorf("Failed to convert signal %s to int", signal)
+		}
+		signum = syscall.Signal(s)
+		// check whether signal is valid or not
+		for _, v := range signals {
+			if v == signum {
+				// signal is a valid signal
+				if err := vc.KillContainer(containerID, podStatus.ContainersStatus[0].ID, signum); err != nil {
```
jodh-intel: I'd prefer only a single call to `vc.KillContainer()` if possible.

Also, note #109 if you want to pick that up?

kill.go
```diff
@@ -129,7 +130,27 @@ func kill(containerID, signal string, all bool) error {
 		return fmt.Errorf("Process not running inside container %s", containerID)
 	}
 
-	if err := vc.KillContainer(containerID, podStatus.ContainersStatus[0].ID, signals[signal]); err != nil {
+	signum, signalOk := signals[signal]
```
dlespiau: You have the same code repeated twice now with some somewhat deep nested indentation.

```
				if err := vc.KillContainer(containerID, podStatus.ContainersStatus[0].ID, signum); err != nil {
					return err
				}
				return nil
```

1. resolve the argument to get a signal number
2. do the KillContainer call

For 1. we don't actually have to cycle through the map, we can just test if the signal is between 1 and 31

kill.go
devimc: yep, I'll take this issue

devimc:
@jodh-intel @sboeuf  @dlespiau  Fixed
Thanks

kill_test.go
```diff
@@ -0,0 +1,63 @@
...
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package main
+
+import (
+	"testing"
+)
+
+func TestProcessSignal(t *testing.T) {
```
dlespiau: The usual thing here to limit the code to type and maintain is to use an anonymous struct: https://dave.cheney.net/2013/06/09/writing-table-driven-tests-in-go

dlespiau: something like

```
tests := []struct {
  input string
  valid bool
  golden
} {
 /* list of test cases */
}

for _, test := range tests {

 .....

}
```

An example at: https://github.com/dlespiau/x86db/blob/master/db_test.go#L10

dlespiau:
with the list of test cases input in a struct this is:

lgtm

kill.go
```diff
@@ -129,9 +130,42 @@ func kill(containerID, signal string, all bool) error {
...
+	signum, signalOk = signals["SIG"+signal]
+	if signalOk {
+		return signum, nil
+	}
+
+	// Support for numeric signals
+	s, err := strconv.Atoi(signal)
+	if err != nil {
+		return 0, fmt.Errorf("Failed to convert signal %s to int", signal)
+	}
```
sboeuf: Add a new line please.

kill.go
```diff
@@ -129,9 +130,42 @@ func kill(containerID, signal string, all bool) error {
...
+	signum = syscall.Signal(s)
+	// check whether signal is valid or not
```
sboeuf: s/check/Check/

kill.go
```diff
@@ -129,9 +130,42 @@ func kill(containerID, signal string, all bool) error {
...
+	for _, v := range signals {
+		if v == signum {
+			// signal is a valid signal
+			return signum, nil
+		}
+	}
```
sboeuf: Add a new line.

kill.go
```diff
@@ -129,9 +130,42 @@ func kill(containerID, signal string, all bool) error {
...
+
+	// Support for numeric signals
+	s, err := strconv.Atoi(signal)
+	if err != nil {
+		return 0, fmt.Errorf("Failed to convert signal %s to int", signal)
+	}
+	signum = syscall.Signal(s)
+	// check whether signal is valid or not
+	for _, v := range signals {
```
sboeuf: Maybe use "sig" instead of "v" because this will be easier to understand.

kill_test.go
sboeuf: Or another way is to actually create each test such as `TestProcessSignalInvalidSignal()`, `TestProcessSignalInvalidShortSignal()`, ... into a generic function like `testProcessSignalFailure()` or `testProcessSignalSuccess()`.
That way each unit test has a name and it is very quick to get which test failed.

sboeuf:
Oh BTW @devimc could you please rebase recent changes from github.com/clearcontainers/runtime:origin/master so that Semaphore CI build can pass.

kill_test.go
dlespiau: @sboeuf that adds quite a lot of typing and duplication actually. If the input needs some description, it can be added to the entries of the table, especially for very simple functions like that one that just take a string as input and return an int.

kill_test.go
```diff
@@ -0,0 +1,41 @@
...
+		{"SIGDCKBY", 0}, //invalid signal
+		{"DCKBY", 0},    //invalid signal
+		{"99999", 0},    //invalid signal
+		{"SIGTERM", syscall.SIGTERM},
+		{"TERM", syscall.SIGTERM},
+		{"15", syscall.SIGTERM},
+	}
+
+	for _, test := range tests {
+		signum, _ := processSignal(test.signal)
```
dlespiau: you're missing testing  we receive a non nil error on invalid input.

devimc:
@dlespiau  @sboeuf  fixed

dlespiau:
\o/

PR #108, opened 2017-05-03
ci: Fix coverage report with travis CI

sboeuf:
Because some permissions were missing, goveralls binary was not installed and we lost the coverage report.


sameo:
LGTM

PR #116, opened 2017-05-05

sboeuf:
This vendoring updates pull the following patches:
df324ac network: Factorize common code between network implementations
9726597 network: Fix wrong MAC address configuration
fcceae7 README: Change shim path in virtc README
7dd546e networking: Turn off multicast snooping on integration bridge
782e907 CONTRIBUTING: Fix the md links

Notice that it fixes the network in two ways
- Disabling multicast snooping on the bridge inside the netns
- Fixing the MAC address provided to the VM

The runtime is now able to ping IP addresses on the host.


sameo:
LGTM

PR #119, opened 2017-05-10
vendor: Update virtcontainers and other packages vendoring

sboeuf:
This PR includes all latest changes for virtcontainers and also updates all packages (except runtime-spec because it has to stay on a specific version).


jodh-intel:
lgtm

PR #129, opened 2017-05-11
Run command

### LINKED ISSUE #94: Implement the runtime run command
`run` creates and starts a container

devimc:
Run command implementation

fixes #94

### DIFF
--- a/run.go
+++ b/run.go
@@ -1,0 +1,98 @@
+// Copyright (c) 2014,2015,2016 Docker, Inc.
+// Copyright (c) 2017 Intel Corporation
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//      http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package main
+
+import (
+	"errors"
+	"fmt"
+	"os"
+
+	"github.com/containers/virtcontainers/pkg/oci"
+	"github.com/urfave/cli"
+)
+
+var runCommand = cli.Command{
+	Name:  "run",
+	Usage: "create and run a container",
+	ArgsUsage: `<container-id>
+
+   <container-id> is your name for the instance of the container that you
+   are starting. The name you provide for the container instance must be unique
+   on your host.`,
+	Description: `The run command creates an instance of a container for a bundle. The bundle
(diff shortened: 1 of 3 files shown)


run.go
```diff
@@ -0,0 +1,94 @@
...
+	},
+}
+
+func run(context *cli.Context) error {
+	runtimeConfig, ok := context.App.Metadata["runtimeConfig"].(oci.RuntimeConfig)
+	if !ok {
+		return errors.New("invalid runtime config")
+	}
+
+	err := Create(context, runtimeConfig)
```
sboeuf: Use
```go
if err := create(); err != nil {
```
instead of
```go
err := create()
if err != nil {
```

run.go
```diff
@@ -0,0 +1,94 @@
...
+	if err != nil {
+		return err
+	}
+
+	pod, err := Start(context.Args().First())
```
sboeuf: s/Start/start/

run.go
```diff
@@ -0,0 +1,94 @@
...
+
+	detach := context.Bool("detach")
+	if detach == false {
+		pid := pod.GetContainers()[0].GetPid()
+		p, err := os.FindProcess(pid)
+		if err != nil {
+			return err
+		}
+
+		_, err = p.Wait()
```
sboeuf: I would retrieve the ProcessState object and print it through the error returned with something like this:
```go
return fmt.Errorf("Process state %s: %s", processState.String(), err)
```

devimc:
@sboeuf fixed

sboeuf:
@devimc cool thanks. You should factorize the flags in common between `run()` and `create()` but this can be done in a follow up PR !
LGTM

run.go
```diff
@@ -0,0 +1,98 @@
+// Copyright (c) 2014,2015,2016 Docker, Inc.
```
sameo: Any reason why we're adding a Docker copyright ?

sboeuf: Good catch, we should remove this, it is coming from the copy/paste from runc I guess.

run.go
```diff
@@ -0,0 +1,98 @@
...
+		return err
+	}
+
+	pod, err := start(context.Args().First())
+	if err != nil {
+		return err
+	}
+
+	detach := context.Bool("detach")
+	if detach == false {
```
dlespiau: `if !detach`. I've seen those comparaisons of bool to true or false literals a bit all over the place and, while I don't have many pet peeves, this one may count as one.

Note that gometalinter will complain about it as well.

run.go
devimc: removing

run.go
devimc: yes, I saw this ```x == false``` in several files, I thought that was ok

sboeuf: Yep that's the way I was testing boolean variable. But I think we have to change that if gometalinter complains about it.

jodh-intel:
lgtm

PR #136, opened 2017-05-16
cli: Allow partial container IDs.

jodh-intel:
Support the specification of partial container IDs. Now, the runtime
will expand a partial container ID as long as The container ID specified:

- is a valid container ID prefix.
- expands to a unique matching container ID.


sboeuf:
LGTM

devimc:
lgtm

PR #137, opened 2017-05-16
cli: Do not load config file when displaying help.

jodh-intel:
Avoid attempting to load the configuration file when the help text
is displayed (when no arguments are specified or when the command is
specified as "help").

Previously, an error would be generated if the system either didn't
have a config file or it was not readable.

### DIFF
--- a/main.go
+++ b/main.go
@@ -96,6 +96,13 @@ func main() {
 	}
 
 	app.Before = func(context *cli.Context) error {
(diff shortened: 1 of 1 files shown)


dlespiau:
lgtm

sboeuf:
LGTM

devimc:
lgtm

PR #140, opened 2017-05-18
Makefile: Install default configuration

amshinde:
The install target should handle installing a default
configuration file as well.
Ran into an issue where the dafaults changed within the code
but the the default configuration installed manually
pointed to old paths.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,5 +1,6 @@
 DESTDIR :=
 BINDIR := /usr/local/bin
+SYSCONFDIR := /etc
 
 SOURCES := $(shell find . 2>&1 | grep -E '.*\.(c|h|go)$$')
 VERSION := ${shell cat ./VERSION}
@@ -7,6 +8,7 @@ COMMIT_NO := $(shell git rev-parse HEAD 2> /dev/null || true)
 COMMIT := $(if $(shell git status --porcelain --untracked-files=no),"${COMMIT_NO}-dirty","${COMMIT_NO}")
 
 TARGET = cc-runtime
+CONFIG = configuration.toml
 
 .DEFAULT: $(TARGET)
 $(TARGET): $(SOURCES)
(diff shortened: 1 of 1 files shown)


sboeuf:
LGTM
BTW, the build is not passing because of the recent proxy changes. Here is the fix: https://github.com/clearcontainers/runtime/pull/141

Makefile
```diff
@@ -23,6 +25,7 @@ check-go-static:
 
 install:
 	$(QUIET_INST)install -D $(TARGET) $(DESTDIR)$(BINDIR)/$(TARGET) || exit 1;
+	$(QUIET_INST)install -D config/$(CONFIG) $(SYSCONFDIR)/clear-containers/$(CONFIG) || exit 1;
```
dlespiau: This should support DESTDIR as well if we're going to support it elsewhere.

Makefile
```diff
@@ -1,12 +1,14 @@
 DESTDIR :=
 BINDIR := /usr/local/bin
```
dlespiau: just a note, it's not part of your PR, but this BINDIR definition should be changed to have a PREFIX. We should be able to specify a PREFIX for our components.

Makefile
amshinde: @dlespiau you mean as in $(DESTDIR)/$(SYSCONFDIR) ?

dlespiau: @amshinde indeed

amshinde: Fixed

Makefile
```diff
@@ -23,6 +25,7 @@ check-go-static:
 
 install:
 	$(QUIET_INST)install -D $(TARGET) $(DESTDIR)$(BINDIR)/$(TARGET) || exit 1;
+	$(QUIET_INST)install -D config/$(CONFIG) $(DESTDIR)$(SYSCONFDIR)/clear-containers/$(CONFIG) || exit 1;
```
dlespiau: sorry, I should have mentioned that earlier, but why the `|| exit `?

amshinde: This was a thoughtless copy from previous line. We dont require it at all. Will take that out.

dlespiau: (mind removing the one on the other line as well? in a different commit if you wish)

dlespiau:
lgtm

jodh-intel:
lgtm

PR #141, opened 2017-05-18
ci: Fix cc-proxy upstart service

sboeuf:
According to latest proxy changes, we cannot use glog flags anymore. Instead we have to call "./cc-proxy --log debug" if we want all debug traces to be provided by the proxy.


sboeuf:
@jodh-intel @devimc please review and merge it so that we can fix the build asap. Thanks!

dlespiau:
lgtm

jodh-intel:
lgtm

PR #143, opened 2017-05-18
create: copy container's annotations to pod

### LINKED ISSUE #133: Run and delete several containers reduce the boot time progressively
```
for i in `seq 0 200 `; do time docker run --rm --runtime=cc30 -ti busybox true  ; done
```

starts with a real time of ```0.802s``` and ends with ```0m5.295s```!
seems like the pod directory is not cleaned and the function ```ListPod()``` takes a long time to return

devimc:
When the pod is deleted we need to know the container's annotations in order to be able to delete the pod.

fixes #133

### DIFF
--- a/create.go
+++ b/create.go
@@ -99,6 +99,11 @@ func create(containerID, bundlePath, console, pidFilePath string,
 
 	process := containers[0].Process()
 
+	// Copy container's annotations.
+	// When the pod is deleted we need to know the container's annotations
+	// in order to be able to delete the pod.
+	pod.SetAnnotations(containers[0].GetAnnotations())
+
 	// config.json provides a cgroups path that has to be used to create "tasks"
 	// and "cgroups.procs" files. Those files have to be filled with a PID, which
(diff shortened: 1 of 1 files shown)


devimc:
depends on https://github.com/containers/virtcontainers/pull/256 and https://github.com/containers/virtcontainers/pull/260

sboeuf:
I think this should be done in the `oci` package provided by `virtcontainers/pkg/oci`. @sameo what do you think ?

dlespiau:
This needs explanations, "how does it fix #133", in the commit message.

sameo:
> I think this should be done in the oci package provided by virtcontainers/pkg/oci

Yes, this should be part of the `oci` package, because this is something all runtime implementations will want.

devimc:
@sameo  @sboeuf  something like this https://github.com/containers/virtcontainers/pull/261 ?

devimc:
move PR to virtcontainers https://github.com/containers/virtcontainers/pull/261

PR #145, opened 2017-05-18
run: delete container's resources once container ends

### LINKED ISSUE #144: runtime standalone: run several containers reduce the boot time progressively
${RUNTIME} --debug --log runtime.log run --pid-file "${BUNDLE}/pid" --bundle "${BUNDLE}" --console $(tty) "${CONTNAME}"

devimc:
fixes #144

### DIFF
--- a/run.go
+++ b/run.go
@@ -82,8 +82,12 @@ func run(context *cli.Context) error {
 
 	detach := context.Bool("detach")
 	if !detach {
-		pid := pod.GetAllContainers()[0].GetPid()
-		p, err := os.FindProcess(pid)
+		containers := pod.GetAllContainers()
+		if len(containers) == 0 {
+			return fmt.Errorf("There are no containers running in the pod: %s", pod.ID())
+		}
+
+		p, err := os.FindProcess(containers[0].GetPid())
 		if err != nil {
 			return err
 		}
@@ -91,6 +95,11 @@ func run(context *cli.Context) error {
 		ps, err := p.Wait()
 		if err != nil {
 			return fmt.Errorf("Process state %s: %s", ps.String(), err)
+		}
+
+		// delete container's resources
+		if err := delete(containers[0].ID(), true); err != nil {
+			return err
 		}
 	}
 


run.go
```diff
@@ -92,6 +92,12 @@ func run(context *cli.Context) error {
 		if err != nil {
 			return fmt.Errorf("Process state %s: %s", ps.String(), err)
 		}
+
+		// delete container's resources
+		err = delete(pod.GetAllContainers()[0].ID(), true)
```
sboeuf: I would prefer:
```go
if err := delete(container.ID(), true); err != nil {
```
And notice that you could factorize upper in the code something like this:
```go
container := pod.GetAllContainers()[0]
```

dlespiau: That code looks like it will panic if GetAllContainers() returns a empty array for some reason.

devimc: that's not possible, if the container is not created ```create``` will fail returning an error

run.go
```diff
@@ -92,6 +92,11 @@ func run(context *cli.Context) error {
 		if err != nil {
 			return fmt.Errorf("Process state %s: %s", ps.String(), err)
 		}
+
+		// delete container's resources
+		if err := delete(pod.GetAllContainers()[0].ID(), true); err != nil {
```
sboeuf: Thanks for the change, but could you do the factorization too please. And the comment from @dlespiau makes sense, it is always better to check we didn't get `nil` before to use a method from this instance.

dlespiau: If we guarantee that GetAllContainers() will indeed return a non empty array with a non nil element after create, not checking sounds fair enough. I wonder if we do guarantee that (not looked).

devimc:
@sboeuf fixed

sboeuf:
Sounds good :)
LGTM

run.go
```diff
@@ -92,6 +96,11 @@ func run(context *cli.Context) error {
 		if err != nil {
 			return fmt.Errorf("Process state %s: %s", ps.String(), err)
 		}
+
+		// delete container's resources
+		if err := delete(containers[0].ID(), true); err != nil {
```
jodh-intel: Good catch!

run.go
```diff
@@ -82,8 +82,12 @@ func run(context *cli.Context) error {
 
 	detach := context.Bool("detach")
 	if !detach {
-		pid := pod.GetAllContainers()[0].GetPid()
-		p, err := os.FindProcess(pid)
+		containers := pod.GetAllContainers()
+		if len(containers) == 0 {
+			return fmt.Errorf("There are not containers running in the pod: %s", pod.ID())
```
jodh-intel: Nit, but could you change "There are not" to "There are no"?

devimc: fixed

jodh-intel:
lgtm

PR #147, opened 2017-05-18
Add CRI-O support

sboeuf:
This PR allows the runtime to be used by CRI-O.


jodh-intel:
Looks good, but there are a lot of new function here, but no new tests to support them.

sboeuf:
@jodh-intel Don't worry, I want to get CRI-O working before I implement the corresponding unit tests ;)

sboeuf:
Don't merge until https://github.com/containers/virtcontainers/pull/268 is merged.

sboeuf:
@jodh-intel PR updated.

sboeuf:
@sameo @jodh-intel we now have the virtcontainers part merged. Let's merge this one ;)

sameo:
LGTM

sameo:
I'd like to get @jodh-intel review though.

Gopkg.toml
```diff
@@ -1,20 +1,107 @@
...
 
-[[dependencies]]
+[[constraint]]
+  branch = "master"
+  name = "github.com/dlespiau/covertool"
+
+[[constraint]]
   branch = "master"
-  name = "github.com/containers/virtcontainers"
+  name = "github.com/docker/docker"
```
jodh-intel: Huh?

delete.go
```diff
@@ -62,58 +62,84 @@ EXAMPLE:
...
 		return err
 	}
 
+	if !force && running {
+		return fmt.Errorf("Container still running, should be stopped")
+	}
+
+	forceStop := false
+	if oci.StateToOCIState(status.State) == oci.StateRunning {
+		forceStop = true
```
jodh-intel: It might be worth adding a log message when this variable is set?

oci.go
```diff
@@ -44,44 +44,64 @@ var (
...
-		if podStatus.ID == containerID {
-			return containerID, nil
+		for _, containerStatus := range podStatus.ContainersStatus {
+			if containerStatus.ID == containerID {
+				return containerStatus, podStatus.ID, nil
+			}
+
+			if strings.HasPrefix(containerStatus.ID, containerID) {
+				if matchFound {
+					return vc.ContainerStatus{}, "", fmt.Errorf("Partial container ID not unique")
```
jodh-intel: Nit: you could use `errors.New()` here.

oci.go
```diff
@@ -245,8 +260,15 @@ func processCgroupsPathForResource(ociSpec oci.CompatOCISpec, resource string) (
 		}
 	}
 
-	if cgroupMountFound == false {
-		return "", fmt.Errorf("cgroupsPath is absolute, cgroup mount MUST exist")
+	if !cgroupMountFound {
+		if isPod {
+			return "", fmt.Errorf("cgroupsPath %q is absolute, cgroup mount MUST exist",
```
jodh-intel: I appreciate this isn't strictly part of this PR, but I find this error rather confusing - the implication is that the path doesn't exist, but what about the absolute bit? Is that a problem too? Could the error become something like this?:

```
"absolute cgroupsPath %q does not exist"
```

state.go
```diff
@@ -68,12 +60,15 @@ func state(containerID string) error {
 	if err != nil {
 		return err
 	}
-	if running == false {
-		if err := stopContainer(podStatus); err != nil {
+
+	if running == false && state.Status == oci.StateRunning {
+		ccLog.Infof("process %d not running, set state as %q",
```
jodh-intel: This message is slightly confusing. Should it be something like this?:

```
"setting container state to %q as process %d is not running"
```

oci.go
sboeuf: I don't think we should modify this error log. This log explains what the OCI spec expects from an absolute path. And this patch does not modify this behavior, it only makes a particular case for CRI-O because there is a bug, they are not handling this OCI specification here. This patch is kind of a workaround to get CRI-O working without modifying the behavior for Docker.

oci.go
sboeuf: Done

state.go
sboeuf: done

oci.go
jodh-intel: Sure, as I said, it isn't really part of your PR, but the current wording just doesn't make sense to me at the moment.

sboeuf: Oh okay got it. But what's wrong about this wording ? If I add
```
"absolute cgroupsPath %q does not exist"
```
This is not really the problem here. Indeed, the OCI spec says that in case of absolute path, we don't want to check if this path exists, but we want to retrieve the mount cgroup which is supposed to be the prefix to concatenate to the cgroup path provided. In that case, we say that we got the following absolute path, but we cannot use it because cgroup mount path is empty.

sboeuf:
@jodh-intel I have no other testing in clearcontainers/tests. I know I have modified the way create/delete/... behave but we had no tests before that, and I still don't have tests for that.

oci.go
jodh-intel: I've created #207 to continue the discussion. I just find the message confusing as it isn't clear which of the conditions is the problem.

jodh-intel:
We have the Semaphore tests (#168) which give us a level of confidence that we haven't broken basic container support. Are there any further tests that can be added there? Otherwise, how are we going to know that CRI-O support is still working?

sameo:
> Otherwise, how are we going to know that CRI-O support is still working?
CRI-O is _not_ working at all without this PR :)

jodh-intel:
@sboeuf - do you have an issue assigned to add CRI-O integration tests?

In my view, we should have more tests on this PR. But if we must consider landing it like this, I think it is only reasonable that you get to fix any "CRI-O bitrot" down the line ;-)

sboeuf:
@jodh-intel I have updated the PR by adding Semaphore CI tests for CRI-O.

jodh-intel:
@sboeuf - excellent, thanks! I appreciate it's all extra work, but I do think this was an essential part of this PR.

lgtm.

chavafg:
@sboeuf why does the CRI-O tests were added here? These tests should have been added in `clearcontainers/tests` repository.

sboeuf:
@chavafg I know those tests should go to clearcontainers/tests, but we wanted this PR to be able to test CRI-O right away, and that was the only way.
But we need those tests properly ported to clearcontainers/tests, I agree.

jodh-intel:
+1 to moving them to the tests repo now, but yes, we had to "bootstrap" these tests into runtime initially.

PR #148, opened 2017-05-18
cli: add config option

devimc:
a configuration file could be specify through --config option
this option could be useful to run functional and integration tests

### DIFF
--- a/main.go
+++ b/main.go
@@ -66,6 +66,10 @@ func main() {
 	app.Version = strings.Join(v, "\n")
 
 	app.Flags = []cli.Flag{
+		cli.StringFlag{
+			Name:  "cc-config",
+			Usage: "clear containers config file path",
+		},
 		cli.BoolFlag{
 			Name:  "debug",
 			Usage: "enable debug output for logging",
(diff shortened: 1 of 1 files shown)


main.go
```diff
@@ -126,7 +130,7 @@ func main() {
 		// Set virtcontainers logger.
 		vc.SetLog(ccLog)
 
-		runtimeConfig, err := loadConfiguration("")
+		runtimeConfig, err := loadConfiguration(context.GlobalString("config"))
```
sboeuf: Just a question here. By default, if we called this binary without any `--config` configuration file provided, the call to `context.GlobalString("config")` will return `""`, right ?

dlespiau:
it may be worth prefixing those cc-specific options so they don't clash with potentially future runc options. Maybe `--cc-options`?

main.go
jodh-intel: Yes: https://godoc.org/github.com/urfave/cli#Context.GlobalString

jodh-intel:
I agree with @dlespiau - naming it `--cc-config=` would be prudent.

devimc:
@dlespiau  @jodh-intel  fixed
thanks

dlespiau:
lgtm

sameo:
LGTM

PR #150, opened 2017-05-19
tests: Fix JSON list format tests for go1.9+.

jodh-intel:
Historically, it has been possible to marshal and demarshal a time.Time and
compare the original and demarshaled values for equality (successfully).

However, go tip (what will presumably become go1.9) changes this behaviour. It
introduces a monotonic element to the time.Time internal representation. Thie
monotonic value (shown as "m=±ddd.nnnnnnnnn" when calling String() on the
time.Time object) is *NOT* retained when marshaling a time.Time.

This ensures comparing the unmarshaled value with the original will fail.

The fix is to call Truncate(0) on the original time.Time object to remove the
monotonic element before attempting comparing with the unmarshaled value.

This change corrects the required "fixup" behaviour in TestStateToJSON() to be
in line with golang expectations.

See: https://go-review.googlesource.com/c/36255/7/src/time/time.go#54

### DIFF
--- a/list_test.go
+++ b/list_test.go
@@ -378,17 +378,27 @@ func TestStateToJSON(t *testing.T) {
 			t.Fatalf("Expected %d lines, got %d", expectedLength, length)
 		}
 
-		// FIXME: Work around bug in golang tip that doesn't maintain
-		// FIXME: the monotonic clock value ("m=±ddd.nnnnnnnnn") when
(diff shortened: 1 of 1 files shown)


dlespiau:
lgtm

sboeuf:
LGTM

devimc:
lgtm

PR #153, opened 2017-05-22
cli: Don't load config file when displaying help for sub-commands.

jodh-intel:
If the user simply want to see a usage statement for a sub-command, the
config file should not be loaded.

Related to #137.


dlespiau:
lgtm

devimc:
lgtm

PR #156, opened 2017-05-22
exec: Rely on oci.CompatOCIProcess instead of specs.Process

sboeuf:
With recent changes from virtcontainers, we have to rely on the compatible structures we have added so that we don't end up with unmarshalling errors regarding `LinuxCapabilities` field.


jodh-intel:
tested fine.

lgtm.

PR #165, opened 2017-05-23
a couple of Makefile tweaks

dlespiau:
Nothing too drastic. Just make sure people can use PREFIX as usual (or almost, we don't have an autogen/configure pass),

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,6 +1,7 @@
 DESTDIR :=
-BINDIR := /usr/local/bin
-SYSCONFDIR := /etc
+PREFIX := /usr/local
+BINDIR := $(PREFIX)/bin
+SYSCONFDIR := $(PREFIX)/etc
 
 SOURCES := $(shell find . 2>&1 | grep -E '.*\.(c|h|go)$$')
 VERSION := ${shell cat ./VERSION}
--- a/Makefile
+++ b/Makefile
@@ -11,7 +11,7 @@ TARGET = cc-runtime
 CONFIG = configuration.toml
 
 .DEFAULT: $(TARGET)
-$(TARGET): $(SOURCES)
+$(TARGET): $(SOURCES) Makefile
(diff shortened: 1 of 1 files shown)


sboeuf:
LGTM

Makefile
```diff
@@ -1,6 +1,7 @@
 DESTDIR :=
-BINDIR := /usr/local/bin
-SYSCONFDIR := /etc
+PREFIX := /usr/local
+BINDIR := $(PREFIX)/bin
```
amshinde: I feel this path should be substituted in the config file we install as well.

jodh-intel: agreed.

dlespiau: Right I was somewhat hoping this would be for another time, but might as might do it properly now. We also need to make sure the runtime is opening the right config file by default, the one in `$(SYSCONFDIR)/clear-containers`.

amshinde: I also prefer to have the default prefix as /usr rather than /usr/local, as we pretty much use that for all our other binaries.

dlespiau: So that's where we have to make a decision. GNU things use /usr/local so a dev version won't override a system version. I've noticed we don't do that, not in cc 2.1, not in shim 3.0 either. I don't care that much but we need to do the same thing everywhere and propose the option to change that.

Installing the config from a dev version to a system one doesn't sound to good.

sboeuf: I would say, let's keep the dev path (/usr/local/...) for now.

jodh-intel: I agree - `/usr/local/` is more trouble for us, but less surprising for other users, so the users win ;)

devimc:
lgtm

sboeuf:
Can we merge this one ?

dlespiau:
I think we can. The functional change is that we now take the prefix into account when installing the config file (so it will be installed in /usr/local/etc by default). There's more work in this area that I haven't captured in issue. Probably a good idea to do it.

PR #169, opened 2017-05-24
run: fix runtime standalone

### LINKED ISSUE #158: failed to run clear container standalone
## command 
```
${RUNTIME} --debug --log runtime.log run --pid-file "${BUNDLE}/pid" --bundle "${BUNDLE}" --console $(tty) "${CONTNAME}"
```

## current output:
```
fork/exec /home/X/gopath/src/github.com/clearcontainers/shim/cc-shim: operation not permitted
```

## expected output:
```
sh-4.3#
```

the issue is that terminal is already the controlling terminal of a different session group, to allow spawned process use assigned terminal we have to steal it form the session group (TIOCSCTTY with arg=1)

### LINKED ISSUE #181: Add support to run Clear Containers standalone
Run command should be able to allocate pseudo terminal in order to be able to use Clear Containers standalone. If ```/dev/ptmx``` or ```/dev/pts/ptmx``` is provided as console argument then the runtime should allocate a pseudo terminal and link it to current terminal

devimc:
With this patch a new pts will be created when the console provided as argument
is /dev/ptmx or /dev/pts/ptmx.

Runtime stdout and stdin will be forwarded to the master fd and then to the
slave, something like this:

```
/dev/pts/0 <--> master <--> /dev/pts/1
```

where pts 0 is assigned to the runtime and pts 1 is assigned to the container.

fixes #181


dlespiau:
eek -4% coverage, that's really a step back and goes against our alpha/beta targets. Will review tomorrow.

console_linux.go
```diff
@@ -0,0 +1,175 @@
...
+
+func (c *linuxConsole) Close() error {
+	if m := c.master; m != nil {
+		return m.Close()
+	}
+	return nil
+}
+
+// mount initializes the console inside the rootfs mounting with the specified mount label
+// and applying the correct ownership of the console.
```
dlespiau: plenty of functions could be pruned form that file, they are used only to handle the console inside a container. That would help the code coverage too.

run.go
```diff
@@ -67,20 +69,49 @@ func run(context *cli.Context) error {
...
 	}
 
-	if err := create(context.Args().First(),
+	consolePath := context.String("console")
+
+	var console Console
+	var err error
+	var consoleState *term.State
+
+	/* if consolePath is /dev/ptmx means that we have to allocate a new pts */
```
dlespiau: /dev/ptmx or /dev/pts/ptmx, actually old-ish runc is uses /dev/pts/ptmx in its tests.

run.go
```diff
@@ -98,9 +129,13 @@ func run(context *cli.Context) error {
 		}
 
 		// delete container's resources
-		if err := delete(containers[0].ID(), true); err != nil {
+		if err = delete(containers[0].ID(), true); err != nil {
 			return err
 		}
+
+		/* close and restore console */
+		console.Close()
```
dlespiau: synchronization is missing there. We need to wait for for goroutine copying console to stdout so we don't miss any bites. That's usually done with a sync.WaitGroup (runc has that). You need to:

- Add a wg.Add(1) just before `go io.Copy(os.Stdout, console)`
- Add a wg.Done() in the go  routine copying console to stdout
- Add a wg.Wait() here.

dlespiau:
For the test coverage, a lot can be done by removing the functions we don't use from `console_linux.go`. Let's see where we are then, but it'd be nice to have some coverage.

dlespiau:
Oh, maybe we should check that the pty is indeed not there when the process ends, at least manually.

Gopkg.lock
```diff
@@ -42,6 +48,11 @@ memo = "3acb5b9124a33a8410d32701c38207f2fcc2c4eec005fd085331a65234904e28"
   version = "v1.1.0"
 
 [[projects]]
+  name = "github.com/docker/docker"
+  packages = ["pkg/term","pkg/term/windows"]
```
devimc: it's not working even in master branch
```
$ dep prune
lock hash doesn't match
```

dlespiau: You could try removing the vendor dir and redoing `dep ensure && dep prune`. I've had that issue before, not sure what it means.

dlespiau:
lgtm

dlespiau:
Of course, latest clear version at this time 15510 doesn't have a -containers.img image.... so tests are failing.

sboeuf:
Please wait for merging this one, I would like to review it properly

run.go
```diff
@@ -67,20 +70,55 @@ func run(context *cli.Context) error {
...
 
+	detach := context.Bool("detach")
+
+	if !detach {
+		wg.Add(1)
+		go io.Copy(console, os.Stdin)
+		go func() {
+			defer wg.Done()
+			io.Copy(os.Stdout, console)
+		}()
```
sboeuf: Why are you not copying stderr ? You should call Add(2) and make sure you're copying stderr the same way you copy stdout, right ?

run.go
```diff
@@ -67,20 +70,55 @@ func run(context *cli.Context) error {
...
+
+		// Save console state because it will be restored once container ends
+		consoleState, err = term.SetRawTerminal(os.Stdin.Fd())
```
sboeuf: Why are we saving only Stdin ?

run.go
```diff
@@ -67,20 +70,55 @@ func run(context *cli.Context) error {
...
+	consolePath := context.String("console")
+
+	var wg sync.WaitGroup
+	var console Console
+	var err error
+	var consoleState *term.State
+
+	// if consolePath is /dev/ptmx or /dev/pts/ptmx
+	// means that we have to allocate a new pts
+	if consolePath == ptmxPath || consolePath == ptsPtmxPath {
```
sboeuf: Are you sure this works in case you use `$(tty)` in your docker command ? If I use `$(tty)`, I always end up with something like /dev/pts/11 where 11 can be any number. How is that supposed to work for this case ?

console_linux.go
```diff
@@ -0,0 +1,133 @@
...
+
+const ptsPtmxPath = "/dev/pts/ptmx"
+
+// ConsoleFromFile creates a console from a file
+func ConsoleFromFile(f *os.File) Console {
+	return &linuxConsole{
+		master: f,
+	}
+}
+
```
sboeuf: We should not copy this entire file, instead we should have only console.go. You can create a structure Console if you want to, but you should rewrite the functions you need, everything happening into console.go.

console.go
```diff
@@ -0,0 +1,32 @@
...
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package main
+
+import (
+	"io"
+	"os"
+)
+
```
sboeuf: As mentioned into console_linux.go, we should have only this file implementing the functions we need, instead of copying all the functions from runc. You can still put a Docker copyright if you feel you are too close from their implementation.

run.go
sboeuf: What about redirecting the signals ?

dlespiau: - we're not copying stdout, stdout is the [destination ](https://golang.org/pkg/io/#Copy)
- we are copying between two terminals, the terminal from which cc-runtime has been launched and the pseudo terminal created for the shim. So onlt two streams of data, the RX/TX of a UART if you wish.
- When terminating, the process has exited so there's no point waiting to flush stdin (but, of course we want to make sure we're flushed the last bytes the process may have written to the terminal).
- There's no signals involved, we're just coying data between to fds.

run.go
dlespiau: We're only saving stdin because:

1. we're modifying stdin (raw mode), so we need to restore it
2. there's no need to do anything on the internal pty the shim will use. If the shim needs to do somethig special to it, it'll do it (and in fact it puts that terminal in raw mode as well).
3. the lifetime of the pty created for the shim stop when `cc-runtime run` exits,

console.go
dlespiau: I mentioned that one (only the functions we need) and it's already addressed.

dlespiau:
I've added some potential follow-up work in issue https://github.com/clearcontainers/runtime/issues/179. I don't think it needs to be part of this PR as it should be useful to start writing tests as is.

run.go
devimc: with $(tty) won't work because in golang you can't steal a tty https://github.com/golang/go/issues/20454

console_linux.go
devimc: console_linux.go only have the functions we need

run.go
dlespiau: We don't support $(tty), it has quite a few not so good consequences we discovered in 2.1.

console_linux.go
dlespiau: I guess @sboeuf is asked to merge console.go and console_linux.go, we don't need the interface at this point.

run.go
sboeuf: Okay fine with me, but do you realize that this PR says it actually fixes https://github.com/clearcontainers/runtime/issues/158 ? And in the description of the bug we are using `$(tty)`, this is confusing.

dlespiau: Yup, I do realise that. Some of the discussion happened on IRC and wasn't logged on issues.

run.go
sboeuf: - Yes you're right, we're not copying stdout. But can't we copy both sterr and stdout fds from the `console` (the one we give to the shim) to the current stdout and stderr from the terminal which launched cc-runtime ?
- Why there is no signal involved ? I don't get how the signals are redirected from the terminal provided in the command line to the actual terminal used by the shim ?

console_linux.go
sboeuf: Yes please, refactor the code.

devimc:
@sboeuf done

sboeuf:
@dlespiau Not sure about the vendoring here. We pull some Windows specific things because docker term depends on this ??
Maybe we can vendor after we took this patch into our runtime codebase ?

dlespiau:
I have talked with @devimc about this and apparently prune doesn't remove it. I guess the docker package pulls that one in.

sboeuf:
LGTM

sboeuf:
@jodh-intel we need your review on this one.

jodh-intel:
lgtm

PR #182, opened 2017-05-26
Sane error for missing pause bundle

jodh-intel:


dlespiau:
and +2.3% coverage, what else can we ask?

lgtm

main_test.go
```diff
@@ -49,3 +49,7 @@ func TestMain(m *testing.M) {
 
 	os.Exit(ret)
 }
+
+func createEmptyFile(path string) (err error) {
```
devimc: Q. Could this function be part of a different file? something like common.go or lib.go?

jodh-intel: I guess it could. Whether we want to do that now or maybe later when we have a few other functions to put in the file I don't know?

dlespiau: Generic files like "common.go", "lib.go", "utils.go" always end up with a mixed bag of "things". Could maaaybe be fileutils_test.go, it's more definite already.

dlespiau: Later is certainly more than fine.

devimc:
lgtm

PR #189, opened 2017-05-29
main: The cc-config is verbose

sameo:
We know it's a clear containers config, it should just be called
--config.

Signed-off-by: Samuel Ortiz <sameo@linux.intel.com>

### DIFF
--- a/main.go
+++ b/main.go
@@ -73,8 +73,8 @@ func main() {
 
 	app.Flags = []cli.Flag{
 		cli.StringFlag{
-			Name:  "cc-config",
-			Usage: "clear containers config file path",
+			Name:  "config",
+			Usage: "clear containers configuration file path",
 		},
 		cli.BoolFlag{
 			Name:  "debug",
@@ -139,7 +139,7 @@ func main() {
 		// Set virtcontainers logger.
 		vc.SetLog(ccLog)
 
-		runtimeConfig, err := loadConfiguration(context.GlobalString("cc-config"))
+		runtimeConfig, err := loadConfiguration(context.GlobalString("config"))
 		if err != nil {
 			return err
 		}


gorozco1:
lgtm

dlespiau:
That was actually on purpose to never clash with runc options.

sameo:
> That was actually on purpose to never clash with runc options.

@dlespiau to avoid getting call  by docker with --config and a runc specific config file ?

jodh-intel:
@sameo - yes, we thought that adding a standard prefix `--cc-*` for all non-standard options we added would make them less likely to class with any future upstream changes in either `runc` or the OCI spec.

dlespiau:
That said, it's quite unlikely it can happen. we could also prefix the option if it happens and not before.

lgtm

jodh-intel:
Weeeeell, possibly. What worries me is if we switch to `--config=` and then runc adds a `--config=` which accepts a different type of config file, users will get confused and their system will be broken unless they are able to update their version of the runtime (which either changes the meaning of `--config=` for us specifically, or does gross things like supporting multiple instances of `--config=` and also handling different data formats for each).

I just think it's safer to namespace now to minimise potential problems for both ourselves and users. If we take a consistent approach to mandate `--cc-*` for our own options and of course document this "special prefix", users will become familiar with any extensions we add and be able to spot them easily.

@devimc - what's your view?

devimc:
> @sameo - yes, we thought that adding a standard prefix --cc-* for all non-standard options we added would make them less likely to class with any future upstream changes in either runc or the OCI spec.

@jodh-intel  I agree with you

devimc:
I remember we had this discussion here https://github.com/clearcontainers/runtime/pull/148

PR #190, opened 2017-05-30
CI: Be verbose when 'gofmt' fails.

jodh-intel:
gofmt is run with the '-s' (simplify) option. This is not the default so
when this fails rather than simply returning a non-zero error code,
display the diff to give the user an indication of what went wrong.


sboeuf:
You should carry this patch to the proxy/virtcontainers as well, and eventually we should carry this script to clearcontainers/tests so that every project relies on the same scripts.

sboeuf:
That being said:
LGTM

devimc:
lgtm

PR #193, opened 2017-05-30
tests: Produce a coverage file with atomic mode

dlespiau:
We're actually using the atomic counters when running our tests because
we run them with -race (2 lines later). We should set the mode of the
resulting coverage file to match this.

Signed-off-by: Damien Lespiau <damien.lespiau@intel.com>


jodh-intel:
lgtm

devimc:
lgtm

PR #194, opened 2017-05-31
Coverage instrumented binary

dlespiau:
The go toolchain doesn't seem to support building coverage instrumented binaries. We can use a [succession of tricks](http://damien.lespiau.name/2017/05/building-and-using-coverage.html) to make it happen.

Fortunately with this, we can very easily meet our 50% coverage target... and aim at 80-90% :)


.ci/go-no-os-exit.sh
```diff
@@ -0,0 +1,17 @@
...
+for f in $candidates; do
+	filename=`basename $f`
+	# skip exit.go where, the only file we should call os.Exit() from.
+	[[ $filename == "exit.go" ]] && continue
+	# skip main_test.go
+	[[ $filename == "main_test.go" ]] && continue
+	files="$f $files"
+done
+
+if grep -n os.Exit $files; then
```
jodh-intel: `if egrep -q '\<os.Exit\>' $files; then` ?

exit.go
```diff
@@ -0,0 +1,32 @@
...
+
+var atexitFuncs []func()
+
+// atexit registers a function f that will be run when exit is called.
+func atexit(f func()) {
+	atexitFuncs = append(atexitFuncs, f)
+}
+
+// exit calls all atexit handlers before exiting the process with status.
+func exit(status int) {
```
jodh-intel: Neat. I wonder if there is any value in printing out the list of registered functions using `reflect` here?

exit.go
```diff
@@ -0,0 +1,32 @@
...
+	for _, f := range atexitFuncs {
```
jodh-intel: If this is really supposed to be like `atexit(3)`, you need to reverse the order so the first function registered gets called last.

.ci/go-no-os-exit.sh
dlespiau: Why -q? we don't want it to be quiet but tell us where offending os.Exit is (file/line)

dlespiau: I guess \. as well

jodh-intel: yep, we need `-n`, not `-q` and should escape the dot as you say.

exit.go
dlespiau: That's a good idea, fixing!

dlespiau:
PR updated:

- Rebased onto master
- exit() calls AtExit() handlers in reverse order
- Be more scripts in the os.Exit check for the CI test

jodh-intel:
lgtm

sboeuf:
lgtm

PR #195, opened 2017-05-31
ci: Add scripts to run cc tests

GabyCT:
This will install the cc-test repository as well as all the
dependencies for the functional tests.

Signed-off-by: Gabriela Cervantes <gabriela.cervantes.tellez@intel.com>

### DIFF
--- a/.ci/install_cc_tests.sh
+++ b/.ci/install_cc_tests.sh
@@ -1,0 +1,21 @@
+# Copyright (c) 2017 Intel Corporation
+
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#      http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+
+#!/bin/bash
+
+set -e
+
+echo "Retrieve cc-tests repository"
+go get -d github.com/clearcontainers/tests
--- a/.ci/install_cc_runtime.sh
+++ b/.ci/install_cc_runtime.sh
@@ -57,6 +57,8 @@ bash .ci/install_cc_shim.sh
 
 bash .ci/install_virtcontainers.sh
 
+bash .ci/install_cc_tests.sh
+
 popd
 
 echo "Start cc-proxy service"


.ci/install_cc_tests.sh
```diff
@@ -0,0 +1,25 @@
...
+#
+
+#!/bin/bash
+
+set -e
+
+echo "Retrieve cc-tests repository"
+go get -d github.com/clearcontainers/tests
+
+echo "Retrieve cc-tests dependencies"
```
dlespiau: This shouldn't be needed, we should vendor all tests dependencies in tests/.

.ci/run_cc_tests.sh
```diff
@@ -0,0 +1,27 @@
+# Copyright (c) 2017 Intel Corporation
```
dlespiau: This is script that will be common to all testing, should be in the `tests` repo.

.ci/install_cc_tests.sh
devimc: but ```ginko``` binary is needed to run our tests

.ci/install_cc_tests.sh
```diff
@@ -0,0 +1,25 @@
...
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+
+#!/bin/bash
+
+set -e
+
+echo "Retrieve cc-tests repository"
+go get -d github.com/clearcontainers/tests
```
jodh-intel: You can remove the `go get`'s below by making this:

```
go get -d github.com/clearcontainers/tests/...
```

.ci/run_cc_tests.sh
devimc: or we can just run ```make functional``` in semaphore

dlespiau: I'm guessing there will be more things at some point, make docker, deal with coverage data, .... May as well have one single script as our "API" and never change that.

.ci/install_cc_tests.sh
dlespiau: ah, hadn't realised that.

.ci/run_cc_tests.sh
GabyCT: @dlespiau ok thanks let me modify it

.ci/install_cc_tests.sh
dlespiau: In any case, this should be part of the `tests` repo. proxy testing will need it as well for instance.

GabyCT:
@dlespiau, changes are updated in this PR and I will do the changes to the ```tests``` repo

.ci/install_cc_tests.sh
GabyCT: @jodh-intel, I tried it in that way however ginkgo is not download it

dlespiau:
There are a few questions left:

- It doesn't seem very clear that we should install the `tests` repo in `install_runtime.sh`. `install_runtime.sh` should just install the runtime, nothing else, I think.
- We don't call into anything in the `tests` repo, so don't run anything atm. Can we all agree to have a script in the `tests` repo all package have to call to run the tests. 

So what about having a script called "run_tests.sh" that will:

1. git clone the `tests` repo
2. call a script in the `tests` repo to run the tests

dlespiau:
Something that could help thinking of a good design for this is thinking about running the tests as an API call or an interface. The `tests` repo provides an "API" to the other packages. One of the calls is "run the tests". We don't need to know about the details of running the tests in this repository, ie, we shouldn't need to install ginkgo here, that's an implementation detail of the `tests` repository.

So now the problem becomes defining a simple API that can be called from runtime, proxy, and shim to setup the test environment and run the tests while sharing the maximum amount of code possible.

chavafg:
Agree with @dlespiau comment above. 
We should call `install_cc_tests.sh` on the project settings on SemaphoreCI. 

But we should also remove these entries:
```
bash .ci/install_cc_proxy.sh

bash .ci/install_cc_shim.sh

bash .ci/install_virtcontainers.sh
```
from `install_runtime.sh` and add them to SemaphoreCI settings.

Thoughts?

jodh-intel:
I'm personally wary of having complex settings for Semaphore since as it doesn't support having a config file like Travis, it is rather too easy to inadvertently remove / re-order scripts using their web interface (I've been bitten by that before).

Although we aren't actually using Semaphore for https://github.com/01org/clr-oci-runtime, I did rework the CI scripts to support it and the intention was always to call *at most* 3 scripts (for any CI) as shown in the Travis config:

https://github.com/01org/cc-oci-runtime/blob/master/.travis.yml#L18..L25

devimc:
I agree with @jodh-intel 
lets keep it simple (KISS) creating 3 scripts 

* setup.sh to install packages and setup the environment
* build.sh or run.sh to test the PR
* cleanup.sh or teardown.sh to report or clean something

currently ```install_runtime.sh``` installs proxy, shim and virtcontainers and it does not make sense

GabyCT:
@dlespiau, @jodh-intel, @devimc and @chavafg, I created this issue https://github.com/clearcontainers/runtime/issues/196 where we will have only 3 scripts in our CI but as this will need a re-work, can we just merged this PR in order to run the functional tests meanwhile?

devimc:
lgtm

dlespiau:
I must be missing something, we won't be running the functional tests with this PR?

GabyCT:
@dlespiau at the clearcontainers/tests I will add in the .ci a script that will run the functional tests so once this PR is merged in semaphore I will run this script (as part of the Jobs).  I need this PR to be merged in order to retrieve the clearcontainers/tests repository.

PR #209, opened 2017-06-12
run: do not duplicate stdio if it is not a terminal

### LINKED ISSUE #208: run command shouldn't duplicate stdin/stdout when they are not present
when a container is launched using the ```run ``` command, the ```runtime``` shouldn't duplicate ```stdin``` and ```stdout``` when  they are not present.
For example when the container is launched by ```ginkgo``` or ```go test```

devimc:
running containers under ginkgo or go test was causing issues
trying to duplicate stdio. with this patch stdio will be
duplicated only if stdio is a terminal

fixes #208

### DIFF
--- a/console.go
+++ b/console.go
@@ -19,6 +19,7 @@ import (
 	"fmt"
 	"io"
 	"os"
+	"syscall"
 	"unsafe"
 
 	"golang.org/x/sys/unix"
@@ -34,6 +35,13 @@ type Console struct {
 
 	master    *os.File
 	slavePath string
+}
+
+// isTerminal returns true if fd is a terminal, else false
+func isTerminal(fd uintptr) bool {
+	var termios syscall.Termios
+	_, _, err := syscall.Syscall(syscall.SYS_IOCTL, fd, syscall.TCGETS, uintptr(unsafe.Pointer(&termios)))
+	return err == 0
 }
 
 // ConsoleFromFile creates a console from a file
--- a/run.go
+++ b/run.go
@@ -85,6 +85,8 @@ func run(context *cli.Context) error {
 		if err != nil {
 			return err
 		}
+		// close and restore console
+		defer console.Close()
 
 		// now consolePath is the slave pts
 		consolePath = console.Path()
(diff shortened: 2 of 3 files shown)


run.go
```diff
@@ -100,7 +101,7 @@ func run(context *cli.Context) error {
 
 	detach := context.Bool("detach")
 
-	if !detach {
+	if !detach && isatty.IsTerminal(os.Stdout.Fd()) {
```
sameo: This is basically:
```Go
func IsTerminal(fd uintptr) bool {
	var termios syscall.Termios
	_, _, err := syscall.Syscall6(syscall.SYS_IOCTL, fd, ioctlReadTermios, uintptr(unsafe.Pointer(&termios)), 0, 0, 0)
	return err == 0
}
```
Could we just copy this implementation rather than pulling yet another dependency ?
["A little copying is better than a little dependency."](https://www.youtube.com/watch?v=PAAkCSZUG1c&t=9m28s)

sameo: Otherwise, thanks for the fix.

devimc:
@sameo fixed

console.go
```diff
@@ -36,6 +37,13 @@ type Console struct {
 	slavePath string
 }
 
+// IsTerminal returns true if fd is a terminal, else false
+func IsTerminal(fd uintptr) bool {
```
sameo: You don't need to export that one, it's reachable inside the main package.

console.go
```diff
@@ -36,6 +37,13 @@ type Console struct {
...
+	var termios syscall.Termios
+	_, _, err := syscall.Syscall6(syscall.SYS_IOCTL, fd, syscall.TCGETS, uintptr(unsafe.Pointer(&termios)), 0, 0, 0)
```
sboeuf: Given the fact that you don't use 4th, 5th and 6th parameters, you should use `syscall.Syscall()` instead of `syscall.Syscall6()`

sameo:
LGTM

sboeuf:
LGTM

PR #214, opened 2017-06-14
cli: Add "cc-check" command.

### LINKED ISSUE #204: Add "cc-check" command
Add a `cc-check` sub-command that will perform the container-specific tests currently provided by [1].

We should not expect users to have to download and run a shell script from a different site to the one hosting the runtime: the runtime should provide the feature itself.

We could simply incorporate these checks into the runtime startup codepath. However, that seems inefficient since the runtime will be called multiple times but the tests only need to be run once (at most). As such, a separate command seems like the best approach initially.

---
[1] - https://download.clearlinux.org/current/clear-linux-check-config.sh

jodh-intel:
This replaces the "clear-linux-check-config.sh" script by providing a
built-in way of determining if the host system is capable of running
Clear Containers.

Fixes #204.

Signed-off-by: Gabriel Briones Sayeg <gabriel.briones.sayeg@intel.com>

### DIFF
--- a/cc-check.go
+++ b/cc-check.go
@@ -1,0 +1,239 @@
+// Copyright (c) 2017 Intel Corporation
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//      http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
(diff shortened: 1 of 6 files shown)


jodh-intel:
Hi @sboeuf - Semaphore is complaining that the CRI-O tests have failed. Any ideas?

jodh-intel:
@sboeuf, @devimc - please can you review?

cc-check.go
```diff
@@ -0,0 +1,239 @@
...
+	}
+
+	return nil
+}
+
+var ccCheckCommand = cli.Command{
+	Name:  "cc-check",
+	Usage: "tests if system can run " + project,
+	Action: func(context *cli.Context) error {
+		err := hostIsClearContainersCapable("/proc/cpuinfo")
```
devimc: should ```"/proc/cpuinfo"``` be a constant ?

devimc:
lgtm

PR #224, opened 2017-06-16
config: Use libexec path rather than hard-coding default shim path.

jodh-intel:
Set the default shim path to be below the prefix set by the Makefile,
rather than just hard-coding it to "/usr/local".


config.go
```diff
@@ -41,11 +41,10 @@ const (
 	pauseBinRelativePath        = "bin/pause"
 )
 
-var defaultShimPath = "/usr/local/libexec/cc-shim"
-
 const (
-	qemuLite = "qemu-lite"
-	qemu     = "qemu"
+	qemuLite   = "qemu-lite"
```
grahamwhaley: guess we'll let you get away with the mixed in whitespace change @jodh-intel  (threw my review eyes out for a moment ;-) )

config.go
```diff
@@ -60,6 +59,10 @@ const (
 	hyperstartAgent = "hyperstart"
 )
 
+// XXX: set by build
```
grahamwhaley: 'XXX' - is that intentional @jodh-intel ?

config.go
```diff
@@ -60,6 +59,10 @@ const (
...
+var libExecDir = ""
```
dlespiau: libexecDir maybe

I think it'd be nice to have sane defaults in this variable so we don't have to check for "" everywhere. NUL value not allowed! :)

grahamwhaley:
minor comments. Technically looks fine to me afaict
lgtm

sameo:
LGTM

PR #227, opened 2017-06-16
Enable checkcommits

jodh-intel:


sboeuf:
@jodh-intel The problem with that PR is that you will get the CI failing if checkcommits fails, right ? In that case, this will prevent the CI from getting further and really test the commit.

jodh-intel:
@sboeuf - yes, the whole point of `checkcommits` is to "fail the PR" if the commits are not in the expected form. But, I suppose we could place the check *after* the CI tests?

Annoyingly, I'm seeing that `checkcommits` is failing for this PR, and yet when I `ssh` into the semaphore env and run it, it works! ;) It could be the tool, but atm it's looking like a possible difference between the "semaphore build environment" and the "semaphore ssh environment".

sboeuf:
@jodh-intel I am fine if you put this after the CI tests, but in case the CI fails for some reasons, you won't check the commit either. I think you should put this in the teardown as it is more an advice than a real failure for the PR.

jodh-intel:
I can see it both ways. `checkcommits` runs in the pre-CI stage for `cc-oci-runtime` as the intention is to enforce its behaviour. The rationale being that if you have submitted a PR with lots of commits and they are not "well-formed", reject the PR as soon as possible to save CI machine time. If we put the check at the end, you are potentially wasting CI machine time as -- the intention is that -- you will need to make your commits "well-formed" and force push. But that will trigger more CI machine time.

To get value out of it, I do think we should "fail" at some point in the CI phase if commits are not well formed as otherwise the tendency will be to ignore its results (who actually looks at their PR CI logs for a successful CI run?)

sboeuf:
Fair enough, leave it as it is right now. I guess we can still decide to change this later if we have some issues with this configuration.

jodh-intel:
An update on the situation...

`checkcommits` works fine under TravisCI, but is not working correctly under `SemaphoreCI` since in the SemaphoreCI build environment [1], there is no way that I can find to determine the range of commits in the PR from the git branch itself. Semaphore provides a variable with the value of `HEAD`, but the `git clone` *only* contains the PR branch (which is called `master`). The real `origin/master` is not available and no references appear to exist to know which commit it's `HEAD` is set to.

I guess "a" way would be to have `checkcommits` do its own `git clone` / `git fetch` of the real `master` branch, but that is not a palatable solution.

I've requested support from Semaphore to see if we can find a better way...

---
[1] - Which, incidentally, appears to be different to the SemaphoreCI ssh environment in this respect (!)

.ci/setup.sh
```diff
@@ -26,6 +26,32 @@ test_repo="github.com/clearcontainers/tests"
...
+set -x
+pwd
+set
+git branch
+git show HEAD
+git show origin
+git reflog
+git status
+
+checkcommits \
```
jcvenegas: Could you add a check to only run it under Travis ?

jodh-intel: Done! ;)

devimc:
lgtm

PR #229, opened 2017-06-16
cc-check: Make the "/proc/cpuinfo" string a const.

jodh-intel:
It is better to define a const for this string since it's safer plus
it is placed near the top of the file for clarity.

### DIFF
--- a/cc-check.go
+++ b/cc-check.go
@@ -33,6 +33,7 @@ type kernelModule struct {
 }
 
 const (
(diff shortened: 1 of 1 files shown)


jodh-intel:
As suggested by @devimc here: https://github.com/clearcontainers/runtime/pull/214#pullrequestreview-44546123.

dlespiau:
lgtm

devimc:
lgtm

PR #232, opened 2017-06-20
cli: Add "cc-env" command

### LINKED ISSUE #228: Add a "cc-env" command
It would be useful to have a `cc-env` command that shows information like:

- kernel being used.
- image being used.
- path to config file (maybe).

jodh-intel:
Introduce a new "cc-env" command that displays configuration details.

Fixes #228.

### DIFF
--- a/cc-env.go
+++ b/cc-env.go
@@ -28,6 +28,18 @@ import (
 	"github.com/urfave/cli"
 )
 
+// Semantic version for the output of the "cc-env" command.
+//
+// XXX: Increment for every change to the output format
+// (meaning any change to the EnvInfo type).
+const formatVersion = "1.0.0"
+
+// MetaInfo stores information on the format of the output itself
+type MetaInfo struct {
+	// output format version
+	Version string
+}
+
 // PathInfo stores a path in its original, and fully resolved forms
 type PathInfo struct {
 	Path     string
@@ -95,9 +107,12 @@ type HostInfo struct {
 	CCCapable bool
 }
 
-// EnvInfo collects all information that will be displayed by the "cc-env"
-// command.
+// EnvInfo collects all information that will be displayed by the
+// "cc-env" command.
+//
+// XXX: Any changes must be coupled with a change to formatVersion.
 type EnvInfo struct {
+	Meta       MetaInfo
 	Runtime    RuntimeInfo
 	Hypervisor PathInfo
 	Image      PathInfo
@@ -106,6 +121,12 @@ type EnvInfo struct {
 	Shim       ShimInfo
 	Agent      AgentInfo
 	Host       HostInfo
+}
+
+func getMetaInfo() MetaInfo {
+	return MetaInfo{
+		Version: formatVersion,
+	}
 }
 
 func getRuntimeInfo(configFile, logFile string, config oci.RuntimeConfig) (RuntimeInfo, error) {
@@ -260,6 +281,8 @@ func getAgentInfo(config oci.RuntimeConfig) (AgentInfo, error) {
 }
 
 func getEnvInfo(configFile, logfilePath string, config oci.RuntimeConfig) (env EnvInfo, err error) {
+	meta := getMetaInfo()
+
 	ccRuntime, err := getRuntimeInfo(configFile, logfilePath, config)
 	if err != nil {
 		return EnvInfo{}, err
@@ -306,6 +329,7 @@ func getEnvInfo(configFile, logfilePath string, config oci.RuntimeConfig) (env E
 	}
 
 	env = EnvInfo{
+		Meta:       meta,
 		Runtime:    ccRuntime,
 		Hypervisor: hypervisor,
 		Image:      image,
--- a/cc-env.go
+++ b/cc-env.go
@@ -1,0 +1,369 @@
+// Copyright (c) 2017 Intel Corporation
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//      http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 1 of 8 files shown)


jodh-intel:
Strictly, this partly fixes #228, but:

- I can't add proxy/shim version details until the following are resolved:
  - https://github.com/clearcontainers/shim/issues/32
  - https://github.com/clearcontainers/proxy/issues/69
  - https://github.com/01org/cc-oci-runtime/issues/981
  - https://github.com/01org/cc-oci-runtime/issues/982
- this PR introduces the framework we can use to add in further details (the above, plus VM detection and maybe dmidecode bits).

utils.go
```diff
@@ -35,3 +48,88 @@ func getFileContents(file string) (string, error) {
...
+func getKernelVersion() (string, error) {
+	contents, err := getFileContents(procVersion)
+	if err != nil {
+		return "", err
+	}
+
+	fields := strings.Fields(contents)
+
+	l := len(fields)
+	if l < 3 {
```
devimc: one line is better ```if len(fields) < 3 ``` :-)

jodh-intel: Good catch - fixing...

cc-env.go
```diff
@@ -0,0 +1,361 @@
...
+
+	vc "github.com/containers/virtcontainers"
+	"github.com/containers/virtcontainers/pkg/oci"
+	specs "github.com/opencontainers/runtime-spec/specs-go"
+	"github.com/urfave/cli"
+	yaml "gopkg.in/yaml.v2"
+)
+
+// CCPath stores a path in its original, and fully resolved forms
+type CCPath struct {
```
dlespiau: FWIW I find the prefixing of those structs with CC a bit ugly, I would either remove it or name them as `CPUInfo` `EnvironmentInfo`, `ShimInfo`, ...

utils_test.go
```diff
@@ -86,3 +86,184 @@ func TestGetFileContents(t *testing.T) {
...
+
+func TestGetCPUDetails(t *testing.T) {
+	type testData struct {
+		contents       string
+		expectedVendor string
+		expectedModel  string
+		expectError    bool
+	}
+
+	validVendorName := "a vendor"
```
dlespiau: A few const could be sprinkled, not a blocker though.

cc-env.go
jodh-intel: Agreed. Fixing...

jodh-intel: Done.

dlespiau: I guess the get* functions could have gone through a s/Details/Info/ as well, anyway that has my lgtm

dlespiau:
lgtm

utils_test.go
jodh-intel: Done.

utils.go
```diff
@@ -35,3 +48,87 @@ func getFileContents(file string) (string, error) {
 
 	return string(bytes), nil
 }
+
+func getKernelVersion() (string, error) {
```
dlespiau: FWIW, I'd just have used the uname syscall (man 2 uname) and https://golang.org/pkg/syscall/#Uname

utils.go
```diff
@@ -35,3 +48,87 @@ func getFileContents(file string) (string, error) {
...
+		contents, err := getFileContents(file)
+		if err != nil {
+			if os.IsNotExist(err) {
+				continue
+			}
+
+			return "", "", err
+		}
+
+		lines := strings.Split(contents, "\n")
```
dlespiau: Something you may be interested in (not necessarily for this PR) is [bufio.Scanner](https://golang.org/pkg/bufio/#Scanner), that would avoid copying the file content a few times (getFileContents, then Split) and directly uses a io.Reader. It's also easier to unit test because one can craft io.Readers backed up by a in-memory slice, say [`strings.Reader`](https://golang.org/pkg/strings/#NewReader)

	scanner := bufio.NewScanner(os.Stdin)
	for scanner.Scan() {
		fmt.Println(scanner.Text()) // Println will add back the final '\n'
	}

utils.go
jodh-intel: I saw that, but I did it manually to ensure we can test it by subverting the `/proc/version` file.

utils.go
jodh-intel: Nice - I'd forgotten about that. I think I'll save that for a follow-up though.

dlespiau: I don't think a follow-up is necessary, not precisely a performance critical piece of code :)

Gopkg.lock
```diff
@@ -120,6 +120,12 @@
   packages = ["unix","windows"]
   revision = "fb4cac33e3196ff7f507ab9b2d2a44b0142f5b5a"
 
+[[projects]]
+  branch = "v2"
+  name = "gopkg.in/yaml.v2"
+  packages = ["."]
+  revision = "cd8b52f8269e0feb286dfeef29f8fe4d5b397e0b"
+
```
sameo: Since we only use yaml to marshal the `ccEnv` structure, couldn't we use json (or toml since we already use it) and not add yet another dependency to our runtime ?

jodh-intel: "yes", but:

- although json is easier to read than XML, it's still not super-easy to read.
- we are admittedly using toml for our config file (I actually don't know where that requirement came from), but I personally find that its embedded tables somewhat odd.

YAML is imho the most human-readable + machine-readable format around atm.

dlespiau: we're just using it to format some output? that can be done without that much code with a few printfs, or even standard library templates (but even that may be overkill)

jodh-intel: This PR did in fact start out using templates but it got slightly unwieldy, hence the simpler marshaling code.

jodh-intel: Realising now how tiny the change is to make `cc-env` output TOML format data instead, I think it would be churlish not to do this.

I'm still much prefer YAML, but -- as you say -- we already have TOML support. One nice touch that TOML has over YAML is that the output is broken into paragraphs ((whereas the YAML gets dumped as a single block of text).

sameo:
LGTM

PR #233, opened 2017-06-21
CI: Quieten CI output.

jodh-intel:
Produce less output by default for some of the CI tests.

### DIFF
--- a/.ci/run.sh
+++ b/.ci/run.sh
@@ -30,24 +30,27 @@ sudo docker exec ${container_id} echo hello
 sudo docker ps -a
 
 # try to unpause a container that isn't paused a few times
+tmpfile=$(mktemp)
 for i in 1 2 3
 do
-	{ sudo docker unpause ${container_id}; ret=$?; } || true
-	[ $ret -eq 1 ]
+	{ sudo docker unpause ${container_id} &>"$tmpfile"; ret=$?; } || true
+	[ $ret -eq 0 ] && { cat "$tmpfile"; exit 1; }
 done
 
 sudo docker pause ${container_id}
-sudo docker inspect ${container_id}|grep -i "\"Paused\":.*true"
+sudo docker inspect ${container_id}|grep -qi "\"Paused\":.*true"
 
 # try to pause an already paused container a few times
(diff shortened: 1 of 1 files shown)


dlespiau:
A fine question: will those script grow enough that we'll need to rewrite this in Go before the end?

lgtm

jodh-intel:
We should atleast `bats` them soon. Is there such a thing as `/bin/gosh` yet? :smile:

sboeuf:
lgtm

.ci/run.sh
```diff
@@ -30,24 +30,27 @@ sudo docker exec ${container_id} echo hello
 sudo docker ps -a
 
 # try to unpause a container that isn't paused a few times
+tmpfile=$(mktemp)
 for i in 1 2 3
 do
-	{ sudo docker unpause ${container_id}; ret=$?; } || true
-	[ $ret -eq 1 ]
+	{ sudo docker unpause ${container_id} >"$tmpfile" 2>&1; ret=$?; } || true
```
devimc: I guess ``` >"$tmpfile" 2>&``` is equal to ```&>"$tmpfile";  ```

.ci/run.sh
```diff
@@ -30,24 +30,27 @@ sudo docker exec ${container_id} echo hello
...
+	[ $ret -eq 1 ] || { cat "$tmpfile"; exit 1; }
```
devimc: Q. does docker always returns ```1``` on failure? wouldn't be better validate against ```0``` ?

devimc: are you expecting ```docker unpause``` fails?

devimc: [ $ret -eq 0 ] && { cat "$tmpfile"; exit 1; }

.ci/run.sh
jodh-intel: Yep - old habits die hard ;) As this is a bash script and not a Bourne shell one, fixed.

.ci/run.sh
jodh-intel: Good catch - fixed.

devimc:
lgtm

PR #240, opened 2017-06-21
exec: Implement detach option

### LINKED ISSUE #318: exec: Implement the detach option
`exec` command does not support the attached mode, by default, everything is run as detached.

sboeuf:
This patch implements the detach option for the exec command. In case
the flag is set to false, we stay attached to the process. This means
we have to wait for the end of the shim process and return the exit
code through the runtime itself.

Docker already implements this option on his own, meaning this patch
is useful for cases where the runtime is used as a standalone binary.

For instance, CRI-O uses the runtime with this option, meaning it
expects this to work as expected.

Fixes #318

### DIFF
--- a/exec.go
+++ b/exec.go
@@ -19,6 +19,8 @@ import (
 	"encoding/json"
 	"fmt"
 	"io/ioutil"
+	"os"
+	"syscall"
 
 	vc "github.com/containers/virtcontainers"
 	"github.com/containers/virtcontainers/pkg/oci"
@@ -220,6 +222,22 @@ func execute(params execParams) error {
 	// after this file is created.
 	if err := createPIDFile(params.pidFile, process.Pid); err != nil {
 		return err
+	}
+
+	if !params.detach {
+		p, err := os.FindProcess(process.Pid)
+		if err != nil {
+			return err
+		}
+
+		ps, err := p.Wait()
+		if err != nil {
+			return fmt.Errorf("Process state %s, container info %+v: %v",
+				ps.String(), status, err)
+		}
+
+		// Exit code has to be forwarded in this case.
+		return cli.NewExitError("", ps.Sys().(syscall.WaitStatus).ExitStatus())
 	}
 
 	return nil


exec.go
```diff
@@ -210,5 +212,20 @@ func execute(params execParams) error {
...
 
+	if !params.detach {
+		p, err := os.FindProcess(process.Pid)
+		if err != nil {
+			return err
+		}
+
+		ps, err := p.Wait()
+		if err != nil {
+			return fmt.Errorf("Process state %s: %v", ps.String(), err)
```
jodh-intel: It would be useful to include details of the container here.

exec.go
```diff
@@ -210,5 +212,20 @@ func execute(params execParams) error {
 		return err
 	}
 
+	if !params.detach {
+		p, err := os.FindProcess(process.Pid)
+		if err != nil {
+			return err
+		}
+
+		ps, err := p.Wait()
```
jodh-intel: This seems to suggest that `exec` is currently broken (fails to wait?)

jodh-intel:
Can you add a test or two to `.ci/run.sh` so we have some level of test coverage for this change please?

exec.go
sboeuf: What do you mean ? No, exec is not broken, we just never implemented the behavior related to detach == false.

exec.go
sboeuf: I'll add that.

exec.go
jodh-intel: Right, currently all `exec`'s are "silently" detached but we never documented that (in the code or in an issue).

sboeuf: Yes exactly, but it was not an issue from Docker perspective because Docker handle this on his own.

jodh-intel:
Hi @sboeuf - any progress on this? We need to find a way to ensure this change is tested somehow.

sboeuf:
Hey @jodh-intel, I have to work on the agent to implement missing things. I'll take care of this after that.

jodh-intel:
Great - thanks @sboeuf!

sboeuf:
@jodh-intel I have rebased the patch on the latest master, but I haven't added any new tests because this has to be done on https://github.com/clearcontainers/tests as mentioned in the issue that I have just opened https://github.com/clearcontainers/tests/issues/171.
Indeed, the standalone tests for the runtime are run with ginkgo. We don't want to duplicate on this repo.

sboeuf:
@dlespiau please take a look.

devimc:
lgtm

sboeuf:
Here is are the tests https://github.com/clearcontainers/tests/pull/180

PR #245, opened 2017-06-23
kill: Don't return an error if container state is 'created'

sboeuf:
Regarding a recent update of the OCI specification, we don't want to generate an error when 'kill' command is called on a container which has been created but not started.

This is basically a no-op in our case.

### DIFF
--- a/kill.go
+++ b/kill.go
@@ -106,8 +106,11 @@ func kill(containerID, signal string, all bool) error {
 
 	containerID = status.ID
 
-	// container MUST be running
-	if status.State.State != vc.StateRunning {
+	// container MUST be created or running
+	if status.State.State == vc.StateReady {
+		ccLog.Infof("Container %s state 'created', nothing to do", containerID)
+		return nil
+	} else if status.State.State != vc.StateRunning {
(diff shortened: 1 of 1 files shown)


sboeuf:
This PR refers to https://github.com/opencontainers/runtime-spec/pull/875

kill.go
```diff
@@ -106,8 +106,11 @@ func kill(containerID, signal string, all bool) error {
 
 	containerID = status.ID
 
-	// container MUST be running
-	if status.State.State != vc.StateRunning {
+	// container MUST be created or running
+	if status.State.State == vc.StateReady {
+		ccLog.Infof("Container state 'created', nothing to do")
```
jodh-intel: Which container?

sboeuf: Updated :)

jodh-intel:
lgtm

amshinde:
lgtm

[Approved with PullApprove](https://pullapprove.com/clearcontainers/runtime/pull-request/245/?utm_source=github-pr&utm_medium=comment-badge&utm_campaign=clearcontainers/runtime)

PR #246, opened 2017-06-26
config: Set default root directory from build.

### LINKED ISSUE #244: Set defaultRootDirectory via the build
#235 missed a path: `defaultRootDirectory` is still hard-coded so needs to be generated like the rest.

jodh-intel:
Generate the default root directory from the build, rather than
hard-coding it.

Fixes #244.


devimc:
lgtm

PR #255, opened 2017-06-27
gitignore: Add generated config files to .gitignore

sboeuf:
We don't want to stage and commit those generated config files by mistake. Then let's add them to .gitignore file.


jodh-intel:
lgtm

dlespiau:
lgtm

PR #259, opened 2017-06-27
kill: Allow 'created' to 'stopped' state transition

sboeuf:
In case the container was created but not running, we were doing
nothing. This patch goes one step further by allowing the update
of the container state when the signal intends to terminate this
container, i.e SIGKILL or SIGTERM.


sameo:
LGTM

PR #270, opened 2017-06-29
Fix cc check coverage

jodh-intel:


cc-check_test.go
```diff
@@ -472,80 +539,118 @@ func TestCheckHostIsClearContainersCapable(t *testing.T) {
...
+	cpuData := []testCPUData{
+		{"GenuineIntel", "lm vmx sse4_1", false},
+	}
+
 	moduleData := []testModuleData{
-		{filepath.Join(sysModuleDir, "kvm"), true, ""},
-		{filepath.Join(sysModuleDir, "kvm_intel"), true, ""},
-		{filepath.Join(sysModuleDir, "kvm_intel/parameters/nested"), false, "Y"},
 		{filepath.Join(sysModuleDir, "kvm_intel/parameters/unrestricted_guest"), false, "Y"},
+		{filepath.Join(sysModuleDir, "kvm_intel/parameters/nested"), false, "Y"},
```
devimc: Q. is nested virtualization needed ?

devimc:
lgtm

cc-check_test.go
jodh-intel: Good question. Technically not, but if it isn't available you won't be able to run a kvm-enabled VM inside a Clear Container (which might be surprising to users0. I guess strictly it should be a warning. #281 raised to deal with that scenario.

PR #272, opened 2017-06-29
build: Ensure user doesn't run "go build".

### LINKED ISSUE #271: stop users calling "go build" directly.
The project needs to be built using `make`, so disallow `go build`.

jodh-intel:
The project must be built using "make", so enforce it.

Fixes #271.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -4,3 +4,4 @@
 /config/configuration.toml
 /pause/pause
 /pause/pause.o
+/error-run-make-not-go-build.go
--- a/Makefile
+++ b/Makefile
@@ -140,6 +140,15 @@ USER_VARS += SHIMPATH
 USER_VARS += SYSCONFDIR
 USER_VARS += PAUSEDESTDIR
 
+# These targets must be in the form "target::" to allow the
+# build check code to be removed and generated for them at the correct
+# point.
+BUILD_CHECK_TARGETS += check
+BUILD_CHECK_TARGETS += check-go-static
+BUILD_CHECK_TARGETS += check-go-test
+BUILD_CHECK_TARGETS += clean
+BUILD_CHECK_TARGETS += coverage
+BUILD_CHECK_TARGETS += $(TARGET)
 
 V              = @
 Q              = $(V:1=)
@@ -185,30 +194,48 @@ endef
 
 export GENERATED_CODE
 
+define BUILD_CHECK_CODE
+package main
+
+// XXX: This file is a sign that you should not run "go build" directly.
+//
+// XXX: Please just run 'make'.
+
+import (
+	"ERROR: use 'go get -d' rather than 'go get'"
+	"ERROR: use 'make' rather than 'go build'"
+)
+endef
+
+export BUILD_CHECK_CODE
+
+BUILD_CHECK_FILE = error-run-make-not-go-build.go
 
 GENERATED_FILES += config-generated.go
 
 config-generated.go: Makefile VERSION
 	$(QUIET_GENERATE)echo "$$GENERATED_CODE" >$@
 
-$(TARGET): $(SOURCES) $(GENERATED_FILES) Makefile | show-summary
+$(TARGET):: remove-build-check-file $(SOURCES) $(GENERATED_FILES) Makefile | show-summary
 	$(QUIET_BUILD)go build -i -o $@ .
 
 pause: pause/pause.o
 	$(QUIET_BUILD)$(CC) -o pause/pause pause/*.o $(CFLAGS) $(LIBS)
 
 .PHONY: \
+	$(BUILD_CHECK_FILE) \
 	check \
 	check-go-static \
 	check-go-test \
 	coverage \
 	default \
 	install \
+	really-clean \
 	show-header \
 	show-summary \
 	show-variables
 
-$(TARGET).coverage: $(SOURCES) $(GENERATED_FILES) Makefile
+$(TARGET).coverage:: $(SOURCES) $(GENERATED_FILES) Makefile
 	$(QUIET_TEST)go test -o $@ -covermode count
 
 $(CONFIG): $(CONFIG_IN)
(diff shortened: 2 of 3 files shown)


jodh-intel:
This PR provides the following behaviour:

```
$ make clean
     CLEAN   clean
$ go build
can't load package: package github.com/clearcontainers/runtime: 
error-run-make-not-go-build.go:8:2: invalid import path: "ERROR: do not call 'go build' directly - just run 'make'!"
$ echo $?
1
$ make
     GENERATE config-generated.go
cc-runtime - version 3.0.0-alpha.2 (commit b17c994786a4232c58f0b5b2581bcc07ca0cd354)

• Summary:

  binary install path (DESTTARGET) : /usr/local/bin/cc-runtime
  config install path (DESTCONFIG) : /usr/local/etc/clear-containers/configuration.toml

     BUILD   cc-runtime
     CONFIG  config/configuration.toml
$ go build
can't load package: package github.com/clearcontainers/runtime: 
error-run-make-not-go-build.go:8:2: invalid import path: "ERROR: do not call 'go build' directly - just run 'make'!"
$ echo $?
1
```

jodh-intel:
This will of course also generate the error on `go get`, hence docs need to refer to `go get -d`.

jodh-intel:
This crazy PR needs more work ;)

jodh-intel:
Updated behaviour (better error message):

```
$ go build
can't load package: package github.com/clearcontainers/runtime:
error-run-make-not-go-build.go:8:2: invalid import path: "ERROR: use 'go get -d' rather than 'go get'"
error-run-make-not-go-build.go:9:2: invalid import path: "ERROR: use 'make' rather than 'go build'"
 echo $?
1
$ make
     GENERATE config-generated.go
cc-runtime - version 3.0.0-alpha.2 (commit a70539cd61c9c2776e8f3675c944f6bffa30fb58)

• Summary:

  binary install path (DESTTARGET) : /usr/local/bin/cc-runtime
  config install path (DESTCONFIG) : /usr/local/etc/clear-containers/configuration.toml

     BUILD   cc-runtime
     CONFIG  config/configuration.toml
cc -Os -Wall -Wextra -static   -c -o pause/pause.o pause/pause.c
     BUILD   pause
$ echo $?
0
$ go build
can't load package: package github.com/clearcontainers/runtime:
error-run-make-not-go-build.go:8:2: invalid import path: "ERROR: use 'go get -d' rather than 'go get'"
error-run-make-not-go-build.go:9:2: invalid import path: "ERROR: use 'make' rather than 'go build'"
$ echo $?
1
```

devimc:
your code look good but how about to use ```config-generated.go``` instead of creating a new file

I just created ```config-generated.go``` with your error message
```
package main

// XXX: This file is a sign that you should not run "go build" directly.
//
// XXX: Please just run 'make'.

import (
	"ERROR: use 'go get -d' rather than 'go get'"
	"ERROR: use 'make' rather than 'go build'"
)
```
and added ```config-generated.go``` to ```.PHONY``` and that's it

```
$ go build
can't load package: package github.com/clearcontainers/runtime: 
config-generated.go:8:2: invalid import path: "ERROR: use 'go get -d' rather than 'go get'"
config-generated.go:9:2: invalid import path: "ERROR: use 'make' rather than 'go build'"
$ echo $?
1
```

```
$ make
     GENERATE config-generated.go
cc-runtime - version 3.0.0-alpha.2 (commit 5c5a5b591974b8c6911600111f34d30da31fcebc-dirty)

• Summary:

  binary install path (DESTTARGET) : /usr/local/bin/cc-runtime
  config install path (DESTCONFIG) : /usr/local/etc/clear-containers/configuration.toml

     BUILD   cc-runtime
$ echo $?
0
```

jodh-intel:
Hi @devimc - I'm not sure I fully understand what you did, but I guess it might make sense to only use a single file as you say.

Maybe we should rename `config-generated.go` to `generated-code.go` to make it clearer that the file isn't just for `config` data, it would then be a "general-purpose" file containing whatever we need it to contain.

@dlespiau, @grahamwhaley - any thoughts?

devimc:
@jodh-intel ```generated-code.go``` sounds better

using the current master branch I just created the file ```config-generated.go``` and added the next content
```
package main

// XXX: This file is a sign that you should not run "go build" directly.
//
// XXX: Please just run 'make'.

import (
	"ERROR: use 'go get -d' rather than 'go get'"
	"ERROR: use 'make' rather than 'go build'"
)
```
and then I added ```config-generated.go``` to the Makefile in the ```.PHONY``` target.

running ```go build``` I get next error
```
can't load package: package github.com/clearcontainers/runtime: 
config-generated.go:8:2: invalid import path: "ERROR: use 'go get -d' rather than 'go get'"
config-generated.go:9:2: invalid import path: "ERROR: use 'make' rather than 'go build'"
```

and that's ok, because we don't want to build the project with ```go build```

jodh-intel:
Ah - I think I've recreated what you've done, although it's not working in all scenarios for me.

That aside, there is a more fundamental problem I think: if the runtime failed at runtime as a result of the generated file code being incorrect, it would be highly confusing for the developer to debug as `generated-code.go` would contain the code to stop the user running `go build` after the build, *not* the config code that `generated-code.go` contained a build time.

jodh-intel:
Semaphore is failing due to https://github.com/clearcontainers/tests/issues/127.

jodh-intel:
@dlespiau, @grahamwhaley - any thoughts on this?

FTR, this is the only way I've found to avoid `go build` (for various repos) doing one of the following when the project has a `Makefile`:

- building seemingly "successfully", but actually building incorrectly (for example, `prog --version` will show incorrect output, but of course it could potentially be a lot more catastrophic than that in some scenarios).
- falling over in a heap (as happens for this repo now).

Another option of course is to close this PR and just document that the user must run `make` - if they don't, they get to deal with the fallout ;) Although this is bordering on a hack, atleast it gives you an indication as to what you are doing wrong.

Once we've decided upon a strategy, we should deploy to all other repos that cannot be built correctly soley using `go build`.

grahamwhaley:
I had one thought (you might not like it.... ;-) - we could place the go bits in a subdir, and have the Makefile at the top level. Thus, a `go build` will do nothing in the top dir (unless you specified a ./... I guess), so it would make it much more obvious you are meant to use the Makefile. Might that be workable?

(I just sort of ran into this with `dep`, which is actually under `cmd/dep`, so confused me why `go install` at the top level was not installing the binary...)
Other than that, I'd have to do some serious digging into `go build` et. al. to see if there were other methods to hand.

jodh-intel:
@grahamwhaley - that sounds like a much cleaner solution to the problem than the contortions on this PR.

@sboeuf, @dlespiau - does this sound good to you? If so, I'll rework this PR to move the code into a subdir.

dlespiau:
Ah, of course, I had forgotten about this.

jodh-intel:
ping @sboeuf + @dlespiau ^ ;)

jodh-intel:
I'm going to close this. Although the source dir trick is neat, on reflection we have documented clearly how to build the runtime here:

- https://github.com/clearcontainers/runtime/blob/master/docs/developers-clear-containers-install.md

Added to which, if you *do* try to build it incorrectly, it will fail to build (and that behaviour is pretty much guaranteed to always be the case now given that the build system generates some of the code).

We can always re-open later...

PR #279, opened 2017-06-29
Vendor virtcontainers and configure Clear Containers specific kernel parameters

sboeuf:
This PR first vendor virtcontainers.
Then it includes the ability to add some kernel parameters to the kernel command line executed by the hypervisor, we have to add Clear Containers specific parameters from the runtime implementation.


sameo:
Tested with a kubeadm deployed k8s: LGTM

PR #285, opened 2017-06-30
config: Give TOML table type variables better names.

### LINKED ISSUE #284: give variables for TOML sections clearer names
Currently, we have variables like this:

```
const ccProxy = "cc"
const ccShim = "cc"
```

The names of those variables only give a vague indication of what they are for. Rename them to make it clearer and add comments!

jodh-intel:
Rename the variables that deal with nested TOML tables and add a
comment explaining what they are for.

Fixes #284.

### DIFF
--- a/config.go
+++ b/config.go
@@ -33,10 +33,6 @@ const (
 	defaultAgent      = vc.HyperstartAgent
 )
 
-const (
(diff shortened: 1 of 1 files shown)


devimc:
lgtm

PR #296, opened 2017-07-04
list: Rename option "--all" to "--cc-all".

### LINKED ISSUE #295: Rename "--all" list option to "--cc-all" for consistency with other extensions.
The list command currently provides a `--all` option. That should be `--cc-all` since this is the standard we have adopted to signify extensions.

jodh-intel:
The list command currently provided a "--all" option, but it should be
called "--cc-all" for consistency with the other extensions.

Also, improve the description of the option and drop the "-a" short-form
(since short options cannot be made Clear Containers specific).

Fixes #295.

### DIFF
--- a/list.go
+++ b/list.go
@@ -104,8 +104,8 @@ To list containers created using a non-default value for "--root":
 			Usage: "display only container IDs",
 		},
 		cli.BoolFlag{
-			Name:  "all, a",
-			Usage: "display all available information",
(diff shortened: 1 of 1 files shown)


devimc:
travis is broken

jodh-intel:
This Travis bug is getting annoying. Let me re-raise...

PR #303, opened 2017-07-06
build: regenerate config-generated.go

### LINKED ISSUE #302: config-generated.go is not regenerated when Makefile and/or VERSION changed
config-generated.go should be regenerated if Makefile and/or VERSION files changed.

dvoytik:
Regenerate config-generated.go if Makefile and/or VERSION files
changed.

Fixes #302.

Signed-off-by: Dmitry Voytik <dmitry.voytik@huawei.com>

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -188,7 +188,7 @@ export GENERATED_CODE
 
 GENERATED_FILES += config-generated.go
 
-config-generated.go:
+config-generated.go: Makefile VERSION
 	$(QUIET_GENERATE)echo "$$GENERATED_CODE" >$@
 
 $(TARGET): $(SOURCES) $(GENERATED_FILES) Makefile | show-summary


dlespiau:
lgtm

dlespiau:
checkcommit vs travis once again. Given the nature of the patch we can merge it as soon as we have a second lgtm. May need to disable checkcommits or find another way to get the list of commits to check.

grahamwhaley:
going to /cc @jodh-intel on it so he can contemplate if this is a checkcommit thing to address etc.

jodh-intel:
lgtm.

@dvoytik - annoying though this is, please could you try force-pushing once more (after forcing a commit hash change). If that doesn't work, I'm afraid you'll have to close and re-open a fresh PR for this :-(

---

Aside: I don't think there is anything `checkcommits` is doing wrong here as it's simply grabbing the value of env vars. We are aware that the tool triggers this bug:

- https://github.com/travis-ci/travis-ci/issues/7830

... but it looks like there is atleast one other Travis bug we're hitting. If anyone can shed light on it, please let us know!

dlespiau:
I think thinking we could make the system more reliable by finding the PR branching point ourselves (this would also mean we have only one code path instead of 1 per CI system, which can be annoying to maintain).

But, for now, let's merge this.

jodh-intel:
@dlespiau - I agree that it would be easier if the tool discovered the branch point itself, but that just isn't possible with the current environment SemaphoreCI provides afaics (well, that's not entirely true, but the tool cannot find the branch point without fetching `master` itself, which is pretty horrid). But I guess we could start with TravisCI nonetheless.

PR #311, opened 2017-07-11
.ci: Increase go test timeout to prevent from false negative tests

### LINKED ISSUE #312: Travis build fails sporadically because go test timeout is too short
Because the go test timeout is too short, and because travis can take some time to run each test sometimes, we run into sporadic failures that we should solve with a bigger timeout.

sboeuf:
Because travis CI can be slow sometimes, the tests can be inconsistent
and fail after a 2 seconds timeout. In order to avoid those behaviors,
this patch increase the timeout from 2 to 5 seconds.

Fixes #312

### DIFF
--- a/.ci/go-test.sh
+++ b/.ci/go-test.sh
@@ -4,7 +4,7 @@ script_dir=$(cd `dirname $0`; pwd)
 root_dir=`dirname $script_dir`
 
 test_packages="."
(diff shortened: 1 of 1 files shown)


dlespiau:
lgtm

grahamwhaley:
lgtm

PR #315, opened 2017-07-11
Implement --console-socket option for compatibility with Docker 17.06

### LINKED ISSUE #293: Support "cc-runtime create --console-socket <path>"
The latest implementation of `runc` support a new `create` command option:

```Shell
   --console-socket value    path to an AF_UNIX socket which will receive a file descriptor referencing the master end of the console's pseudoterminal
```

We need to support this option to have CC 3.0 work with containerd 1.0 as it relies on it for all interactive containers.

### LINKED ISSUE #316: Support "cc-runtime exec --console-socket <path>"
This new option has been introduced by the latest version of `runc` and we want to support it in order to be compatible with Docker 17.06.

### LINKED ISSUE #317: Support "cc-runtime run --console-socket <path>"
This new option has been introduced by the latest version of `runc` and we want to support it in order to be compatible with Docker 17.06.

sboeuf:
Docker 17.06 relies on this new option because it relies on the latest version of runc (1.0.0-rc3). This PR implements this option for `create`, `run` and `exec`.

Fixes #293 
Fixes #316 
Fixes #317


devimc:
lgtm

amshinde:
lgtm

dvoytik:
Would it be easier to copy & paste https://github.com/opencontainers/runc/blob/master/libcontainer/utils/cmsg.go instead of vendoring 11K of lines?

sboeuf:
@dvoytik I agree with that but the reason why I made this choice is because I want to stay in sync with what is defined by runc. This means that if the SendFd functions changes for runc, then Docker will expect something else and we will break our compatibility.

PR #324, opened 2017-07-13
exec: Override container process option if needed

### LINKED ISSUE #323: exec: Use container process specs and override if necessary
When doing `exec`, we only rely on the option passed through the command line to setup the process environment. But `runc` expects that we use the options defined when the container was started, and to override them if necessary.

sboeuf:
runc expects the exec command to use the same options defined when
the container process was started. And if the command line provide
some options, we should override with what is provided.

Fixes #323

### DIFF
--- a/exec.go
+++ b/exec.go
@@ -114,16 +114,11 @@ EXAMPLE:
 		},
 	},
 	Action: func(context *cli.Context) error {
-		params, err := generateExecParams(context)
-		if err != nil {
-			return err
-		}
-
-		return execute(params)
+		return execute(context)
 	},
 }
 
-func generateExecParams(context *cli.Context) (execParams, error) {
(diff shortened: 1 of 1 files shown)


exec.go
```diff
@@ -150,23 +145,58 @@ func generateExecParams(context *cli.Context) (execParams, error) {
...
+
+		// Override user
+		if context.String("user") != "" {
+			params.ociProcess.User = specs.User{
+				Username: context.String("user"),
+			}
+		}
+
+		// Override env
+		params.ociProcess.Env = append(params.ociProcess.Env, context.StringSlice("env")...)
```
devimc: what happen if the len of context.StringSlice("env") is 0?

sboeuf: I think it is a no-op in that case.

devimc:
lgtm

PR #327, opened 2017-07-14
console: Don't link the shell to the shim if detach and tty are enabled

### LINKED ISSUE #326: Standalone and detach=true/tty=true command (run, exec) should not output anything to the shell
In case our runtime is used in standalone mode with the `--detach` and `--tty` flags enabled, we should not get any output on the current shell. In any case, we should still be able to get some outputs if a console has been provided.

sboeuf:
In case our runtime is used in standalone mode with the --detach
and tty flags enabled, we should not get any output on the current
shell. We should still be able to get some outputs if a console has
been provided though.

This patch relies on the latest virtcontainers vendoring to be
able to indicate if we are in case of a detached command or not.

Fixes #326

### DIFF
--- a/Gopkg.lock
+++ b/Gopkg.lock
@@ -41,7 +41,7 @@
   branch = "master"
   name = "github.com/containers/virtcontainers"
   packages = [".","pkg/cni","pkg/hyperstart","pkg/oci"]
-  revision = "e01ebd015f77b3e822b597088e6efc102a7e6d9a"
+  revision = "cfe8504aee2d22812dba5d8de17e545359c4a2fa"
 
 [[projects]]
   branch = "master"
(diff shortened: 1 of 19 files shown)


devimc:
lgtm

jodh-intel:
One of the commits is missing a subsystem prefix, but aside from that...

lgtm

sboeuf:
@jodh-intel I know the third commit is missing a subsystem prefix but I didn't know what to use. Any idea ?

jodh-intel:
@sboeuf - it's not really rigid: as long as it is descriptive so maybe one of `tty`, `console`, `terminal`, `output` or something like that.

sboeuf:
@jodh-intel ok sounds good. I was trying to tie this to something like `run`, `exec` but I think your proposal is much better :)

sboeuf:
@jodh-intel please merge this if this is fine for you :)

PR #337, opened 2017-07-18
vendor: Update github.com/BurntSushi/toml for license update

### LINKED ISSUE #335: vendoring: update BurntSushi/toml to get new license
The toml library we use updated to a new license (that fits us better) a little while back:
https://github.com/BurntSushi/toml/pull/184
re-vendor that package to pick it up.
https://github.com/clearcontainers/runtime/blob/master/vendor/github.com/BurntSushi/toml/COPYING

sboeuf:
The most recent code from github.com/BurntSushi/toml moved from the WTF license to the MIT license.

Fixes #335


Gopkg.lock
```diff
@@ -9,15 +9,9 @@
 
 [[projects]]
   branch = "master"
-  name = "github.com/Azure/go-ansiterm"
```
grahamwhaley: Just to check - we understand and wanted this Azure delete at the same time?
I did a `dep ensure -update github.com/BurntSushi/toml` - and understand the numerous other changes that happen to the lock file are just part of what `dep` did (and I half consider that maybe a `dep` bug - we only asked to update a single package, but it went and did a bunch of other stuff as well, and afaik there is no way to tell it not to....).

Gopkg.lock
```diff
@@ -63,12 +57,6 @@
 
 [[projects]]
   branch = "master"
-  name = "github.com/docker/docker"
```
grahamwhaley: And, to check, we understand why docker got dropped

Gopkg.lock
sboeuf: @grahamwhaley this is not a bug, this is because we removed recently a part of code relying on this docker package. And when you ask dep to do the dep ensure, it first try to make sure that everything is right on the repo. Thus, we are using this update of toml to update properly our repo vendoring.

sboeuf: But I have split this PR in two separate commits on purpose, so that we don't hide this in only one commit.

grahamwhaley: ah, didn't spot the commit split - nice move.
re: is it a bug - well, I guess that's up to the dep authors, but seeing as dep was given a single explicit instruction (update this one package), and it wen't and did some other (admittedly not 'incorrect breakage') stuff - really I'd like them to have an option so you have to knowingly 'force' the updates in this case or something. If I find a minute I'll trawl the dep code/github to see if they've covered this before etc.

lgtm

grahamwhaley:
lgtm
coz it looks like my in-line one didn't trigger....

sboeuf:
@grahamwhaley you can do what you describe by doing in this specific order:
```
dep ensure
dep prune
```
This will give you what changed in case you removed or added a new dependency
And then
```
dep ensure -update github.com/BurntSushi/toml
dep prune
```
will update only the `toml` package.

PR #352, opened 2017-07-20
cc-env: Add machine type to hypervisor information

### LINKED ISSUE #351: cc-env: Add hypervisor machine type info
cc-env needs to provide the hypervisor machine type through the information provided.

sboeuf:
The purpose of this commit is to add some useful info related to the
hypervisor now that we have the machine type available.

EnvInfo structure has been modified so that it expects HypervisorInfo
structure instead of the generic PathInfo regarding the information
about the hypervisor.

Fixes #351


devimc:
lgtm

PR #353, opened 2017-07-20
runtime: Switch to the new agent

### LINKED ISSUE #348: Switch to the new agent
Test the new agent and validate that we cannot see any regression. Once this is done, let's move to it.

sboeuf:
This commit modifies the systemd target to use, so that it will use
cc-agent instead of hyperstart agent.

Fixes #348

### DIFF
--- a/create.go
+++ b/create.go
@@ -156,7 +156,7 @@ func createPod(ociSpec oci.CompatOCISpec, runtimeConfig oci.RuntimeConfig,
 		},
 		{
 			Key:   "systemd.unit",
-			Value: "cc-agent.target",
+			Value: "clear-containers.target",
 		},
 		{
 			Key:   "systemd.mask",
--- a/versions.txt
+++ b/versions.txt
@@ -1,0 +1,1 @@
+cc_agent_version=0fca1509afbaa18c5a0ddf213f2e377c7b87dcc7


jcvenegas:
Given that it depends on a cc image version, don't we need something like ’versions.txt‘?

sboeuf:
@jcvenegas I only wanted to make a basic check but this is not working for an unknown reason. Maybe the kernel we use in clearcontainers/tests is too old.

jodh-intel:
Please can you add details to the https://github.com/clearcontainers/runtime#debugging explaining how to view the agent log.

sboeuf:
@jcvenegas yes that's a good idea, I have introduced versions.txt mentioning our agent version commit.

sboeuf:
@jodh-intel Changing the agent won't change the way we get the logs. We will get them from through the proxy, the same way we are doing with hyperstart currently.

jodh-intel:
@sboeuf - right, but I don't think we've documented anywhere how to actually view the agent log at all(?) Since this is new code, I'd say there is a good chance we might need to see the log and potentially compare with hyperstart. You are right, it isn't strictly required for this PR, but it would be useful. If not on this PR, I've raised #366.

gorozco1:
Any plans to merge this today?

gorozco1:
@erick0z will build a `cc-runtime` with this patch and place it under  http://download.opensuse.org/repositories/home:/clearcontainers:/clear-containers-3-staging/  for @chavafg to test.

sboeuf:
@jodh-intel @mcastelino @grahamwhaley or @devimc Could you give a LGTM to this PR so that we can merge it ?
High priority for BETA release before EOD.

devimc:
lgtm

Issue #382, opened 2017-08-04
vendor: Update virtcontainers vendoring


sboeuf:
Needs to update virtcontainers vendoring to get latest patches from virtcontainers.

Issue #378, opened 2017-08-01
kill: Simplify kill command after recent virtcontainers changes


sboeuf:
Because virtcontainers is now re-organized so that it can handle the different cases according to container states, we want to simplify the implementation of the kill command to avoid any confusion between the runtime and virtcontainers.

Issue #377, opened 2017-08-01
delete: Always stop the pod from delete command


sboeuf:
In case of a container, if kill has been previously called and has terminated the container's process, we now have virtcontainers updating the container state (https://github.com/containers/virtcontainers/pull/333 and https://github.com/containers/virtcontainers/pull/327). But the pod state will never switch to "stopped" implicitly. We have to stop the pod on purpose to end properly the lifecycle. This means in case delete command is called for the pod itself (Docker or CRI-O pod cases), we have to call into StopPod() no matter what is the value of the force flag.

PR #383, opened 2017-08-04
Globally improves kill/delete commands

### LINKED ISSUE #376: Don't check for shim status since virtcontainers do that already
There is a pending PR (https://github.com/containers/virtcontainers/pull/333)/ issue (https://github.com/containers/virtcontainers/issues/327) for virtcontainers to be able to update the container status when StatusContainer() is called. This means the runtime implementation does not need to check the shim state on purpose, it can now rely on the state provided by virtcontainers.

sboeuf:
This PR includes recent changes from virtcontainers. It changes the way `delete` and `kill` commands are implemented and allows `cc-runtime list` to return an accurate result of the container status.
It fixes #376 #377 #378 and #382

### DIFF
--- a/delete.go
+++ b/delete.go
@@ -80,19 +80,13 @@ func delete(containerID string, force bool) error {
 		return err
 	}
 
-	running, err := processRunning(status.PID)
-	if err != nil {
-		return err
-	}
-
-	if !force && running {
-		return fmt.Errorf("Container still running, should be stopped")
-	}
-
 	forceStop := false
 	if oci.StateToOCIState(status.State) == oci.StateRunning {
+		if !force {
+			return fmt.Errorf("Container still running, should be stopped")
+		}
+
 		forceStop = true
-		ccLog.Info("Force stopping the pod/container before deleting")
 	}
 
 	switch containerType {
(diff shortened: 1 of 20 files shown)


sboeuf:
@devimc @amshinde @mcastelino could you please take a look ?

delete.go
```diff
@@ -113,11 +113,9 @@ func delete(containerID string, force bool) error {
 	return removeCgroupsPath(cgroupsPathList)
 }
 
-func deletePod(podID string, forceStop bool) error {
-	if forceStop {
-		if _, err := vc.StopPod(podID); err != nil {
-			return err
-		}
+func deletePod(podID string) error {
```
mcastelino: @sboeuf are you supposed to allow the delete of a running container.

sboeuf: No we should not be able to delete a running container, and this is checked by the code here https://github.com/clearcontainers/runtime/pull/383/files/618ab6482064bc977f802abed9084514e5569d45..4dc190d917a7eb21b00db85a0e0c0cf746979a1e#diff-635f3cde55281671a8db5283d1597a71R84

mcastelino: LGTM

amshinde:
lgtm

PR #408, opened 2017-08-11
qemu: Move from qemu-lite to qemu-cc binary

### LINKED ISSUE #407: qemu: Move from qemu-lite to qemu-cc

sboeuf:
Because we want to run Clear Containers on the most recent and
maintained version of qemu-lite, we are going to rely on qemu-cc
binary which is the packaged version of qemu-lite-q35 (or most
commonly known as 2.9.0).

Basically we are moving from qemu-lite 2.7 to 2.9.

Fixes #407

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -73,11 +73,11 @@ IMAGEPATH := $(PKGDATADIR)/clear-containers.img
 
 KERNELPARAMS :=
 
-# The CentOS/RHEL hypervisor binary is not called qemu-lite
-ifeq (,$(filter-out centos rhel,$(distro)))
+# The RHEL hypervisor binary is not called qemu-cc
+ifeq (,$(filter-out rhel,$(distro)))
 QEMUCMD := qemu-system-x86_64
 else
-QEMUCMD := qemu-lite-system-x86_64
+QEMUCMD := qemu-cc-system-x86_64
 endif
 
 QEMUPATH := $(QEMUBINDIR)/$(QEMUCMD)


sboeuf:
@grahamwhaley @jodh-intel @mcastelino @devimc we need an approval on this one for when Semaphore will be able to pass because everything else will be merged.

grahamwhaley:
lgtm
Just for completeness, the new qemu build comes from the tree here: https://github.com/clearcontainers/qemu/tree/qemu-lite-v2.9.0

We got an odd Semaphore failure (system power down) - I'll see if a 'label wiggle' will restart it...

grahamwhaley:
ah, there is a hint that maybe @sboeuf stopped the semaphore build (or the logs could be misleading). Anyway, I'll let @sboeuf rekick semaphore as and when necessary

Makefile
```diff
@@ -77,7 +77,7 @@ KERNELPARAMS :=
 ifeq (,$(filter-out centos rhel,$(distro)))
```
gorozco1: We have packages for `qemu-cc` too, need to be added to the test, this conditional should be only for rhel IMHO

sboeuf: Oh ok, let me change that!

Makefile
```diff
@@ -74,10 +74,10 @@ IMAGEPATH := $(PKGDATADIR)/clear-containers.img
 KERNELPARAMS :=
 
 # The CentOS/RHEL hypervisor binary is not called qemu-lite
```
gorozco1: Just one more change, please remove CentOS from the comment and rename `qemu-lite` to `qemu-cc`

sboeuf:
@gorozco1 just updated it :) Thanks for the review

chavafg:
After running tests with this new qemu-cc some cri-o failures appeared. More information on https://github.com/clearcontainers/tests/issues/364 . 
Closing this PR for the moment as we are not moving with qemu-cc until all problems are resolved.

PR #422, opened 2017-08-14
cc-check: Document "unrestricted_guest" requirement

### LINKED ISSUE #421: Document why "unrestricted_guest" is checked for
`cc-check` requires the KVM feature `unrestricted_guest`. This is a heuristic used to ensure the host system is "new enough" to run a Clear Container.

See: https://github.com/01org/cc-oci-runtime/issues/950#issuecomment-307148142

jodh-intel:
The `cc-check` command requires the "unrestricted_guest" KVM feature.
This is used as a heuristic to determine if the system is new enough to
run a Clear Container.

Fixes #421.

### DIFF
--- a/cc-check.go
+++ b/cc-check.go
@@ -86,7 +86,11 @@ var requiredKernelModules = map[string]kernelModule{
 	"kvm_intel": {
 		desc: "Intel KVM",
 		parameters: map[string]string{
-			"nested":             "Y",
+			"nested": "Y",
+			// "VMX Unrestricted mode support". This is used
+			// as a heuristic to determine if the system is
+			// "new enough" to run a Clear Container
+			// (atleast a Westmere).
 			"unrestricted_guest": "Y",
 		},
 	},


cc-check.go
```diff
@@ -69,7 +69,11 @@ var requiredKernelModules = map[string]kernelModule{
 	"kvm_intel": {
 		desc: "Intel KVM",
 		parameters: map[string]string{
-			"nested":             "Y",
+			"nested": "Y",
+			// "VMX Unrestricted mode support". This is used
```
grahamwhaley: I'd like to have @mcastelino and @anthonyzxu verify/clarify this. I think there is a more subtle set of interactions to do with `unrestricted_mode` (and if our QEMU supports software emulation of that if the hardware does not have the feature), and QEMU `disable-modern=true`, which changes the virtio PCIe IO/MMIO functionality. I expect there is a tight bond between using `modern MMIO` virtio PCIe devices and having `unrestricted_mode` supported or emulated in QEMU.

In summary, I think we can be clearer here - and maybe even note that if the feature is not present you can enable CC by modifying the hypervisor.args (QEMU command line - or maybe we add a config option for CC3.x runtime).

mcastelino: @grahamwhaley unrestricted mode is used by KVM to directly handle execution of the VM when it is in real and protected mode. We need this to eliminate the use of emulation. There is no co-relation between this and virtio 1.0. We do not have support for emulation in q35-lite and pc-lite. So checking for this is still mandatory.

sboeuf: @mcastelino that's weird cause Azure VM does not have this option enabled and I have been able to get CC3.0 running with our current setup (qemu-lite with pc machine type) + `disable-modern=true` for virtio devices.

mcastelino: @sboeuf that is unexpected. Let me look at that. We had eliminated this code at least for pc-lite.

sboeuf: Ok, keep me posted about that.

jodh-intel: Reading the comments above, it's unclear to me how the comment should be reworded given the Azure angle. Any suggestions folk?

sboeuf:
@jodh-intel https://github.com/containers/virtcontainers/pull/343 is now merged, you can rely on the function `RunningOnVMM()` from `hypervisor.go` to detect if running in nested environment and take the decision about `unrestricted_guest`

cc-check.go
jodh-intel: pingety, ping! Any word-smiths out there? :smile:

grahamwhaley: I think we need slightly more tech clarification here before we wordsmith it - I'm still not 100% clear on all the situations with `unrestricted_guest` and nested VMs.
We worked out the other bits - that we needed to revert to virtio 0.9 and we had to remove pmu or it tried to access some MSRs that did not exist in the VM - but, I cannot quite fathom where we ended up with the relationship of needing `unrestricted_guest` and nested VMs.
Over to @mcastelino and @sboeuf to clarify that I think (sorry guys - I know you are probably a bit fed up with that loop now)

anthonyzxu: Graham,
I don't think we need to check unrestricted_guest.
Qemu-lite only run guest in protected paging mode.
Qemu-lite should work in machine which doesn't have unrestricted_guest feature.

jodh-intel:
@mcastelino - can we remove this check entirely now then?

cc-check.go
jodh-intel: Hi @anthonyzxu - understood. But I believe we use this check as a heuristic to determine if the system is "new enough". Do we believe there could be users that will be trying to run on a system without this feature? If not, great, but if so, we need to find another heuristic to make `cc-check` more reliable.

It's worth pointing out that iff `unrestricted_guest` is not available, `cc-check` will only warn, not error. And of course, `cc-check` is not called by the runtime under a container manager so users will not be stopped from running if this feature is not available.

/cc @sameo, @grahamwhaley, @mcastelino.

jodh-intel:
We don't seem to be able to reach a decision on whether this check is needed.

However, can we atleast agree to land this PR, which merely adds a descriptive comment explaining why we currently perform the check?

At this rate, this PR is going to be on solid foods soon... :smile:

grahamwhaley:
Agreed, as a comment update,
lgtm

but all the CI's look bust - I guess it might just need a mega rebase ?

jodh-intel:
/me blows the dust off the branch, rebases and re-pushes...

PR #424, opened 2017-08-14
tests: Add framework to mock virtcontainers

### LINKED ISSUE #401: increase unit test coverage on create.go
Currently very low coverage.

See #93.

jodh-intel:
Add a simple test framework to allow the behaviour of the virtcontainers
package to be mocked.

This change introduces a RVC interface and provides two implementations:
one for virtcontainers itself (rvc-implementation.go) and another mock
implementation for the tests to allow the behaviour of the
virtcontainers package to be modified (rvc-implementation_test.go). By
only accessing the virtcontainers implementation via the "vci" variable,
the tests can be switched to the mock implementation.

Note that this mocking is used rather than testing virtcontainers
directly since the virtcontainers API is already tested in that package:
the mock framework thus avoids duplication of test code and allows more
fine-grained error scenarios to be handled.

This change also introduces a single new test, TestCreate(), that
increases the unit test coverage of create.go by using the new framework
to force create() to fail due to a virtcontainers failure in
CreatePod().

Fixes #401.

### DIFF
--- a/rvc-implementation.go
+++ b/rvc-implementation.go
@@ -1,0 +1,98 @@
+// Copyright (c) 2017 Intel Corporation
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//      http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+// Description: The true virtcontainers implementation of the RVC interface.
+// This indirection is required to allow an alternative implemenation to be
+// used for testing purposes.
+
+package main
+
+import (
+	"syscall"
+
+	"github.com/Sirupsen/logrus"
+	vc "github.com/containers/virtcontainers"
+)
+
+// virtcontainers implementation type
+type vcImpl struct {
+}
+
+func (impl *vcImpl) SetLogger(logger logrus.FieldLogger) {
+	vc.SetLogger(logger)
+}
+
+func (impl *vcImpl) CreatePod(podConfig vc.PodConfig) (*vc.Pod, error) {
+	return vc.CreatePod(podConfig)
+}
+
+func (impl *vcImpl) DeletePod(podID string) (*vc.Pod, error) {
+	return vc.DeletePod(podID)
+}
+
+func (impl *vcImpl) StartPod(podID string) (*vc.Pod, error) {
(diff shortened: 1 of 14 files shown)


jodh-intel:
Once this has landed, I'll follow-up with:

- More unit tests for `create.go`.
- A PR to also mock the virtcontainers `oci` package.

rvc-implementation.go
```diff
@@ -0,0 +1,98 @@
...
+	"github.com/Sirupsen/logrus"
+	vc "github.com/containers/virtcontainers"
+)
+
+// virtcontainers implementation type
+type vcImpl struct {
+}
+
+func (impl *vcImpl) SetLogger(logger logrus.FieldLogger) {
+	vc.SetLogger(logger)
```
grahamwhaley: fmi, are we expecting any performance impact at all with the indirection, or do we think either Go will optimise it away, or it will be so small and we make so few actual calls and only at startup that it will be negligible?

rvc-implementation_test.go
```diff
@@ -0,0 +1,278 @@
...
+
+	vc.SetLogger(logger)
+}
+
+func (impl *mockImpl) CreatePod(podConfig vc.PodConfig) (*vc.Pod, error) {
+	if impl.createPodFunc != nil {
+		return impl.createPodFunc(podConfig)
+	}
+
+	if impl.forceFailure {
```
grahamwhaley: Just curious, I'm presuming the decision to call the underlying impl if it was registered before checking if we should return error was deliberate. Question then, might there be times when we want to return an error without calling the underlying function (in which case did you intend for folks to set impl.xxx to nil before they do the call, maybe?)

rvc-implementation.go
jodh-intel: I think the impact should be negligible but what do you think @sboeuf ?

grahamwhaley:
@jodh-intel as you have a CI niggle, I'm going to mark this WIP/DNM right now, and you can clear that down again when you are happy we are good to go...

rvc-implementation.go
amshinde: @jodh-intel I agree with sboeuf's comment below, why not have the mock implementation in virtcontainers package?  That way we do not need to introduce the indirections which add yet another call and also make it a little less intuitive with code calling into vc for the state constants and vci for the actual lifecycle calls.

grahamwhaley:
I have to say, @sboeuf got me hooked with `can be reused by any other consumer of the virtcontainers library`. I like that idea - @jodh-intel what do you think, can you see any reasons this could not be done/work over in virtcontainers - anything harder, miss any functionality, more duplication etc. etc.?

jodh-intel:
I agree with @sboeuf and am currently investigating integrating something like this PR into virtcontainers.

jodh-intel:
Closing in favour of introducing this framework in virtcontainers under:

- https://github.com/containers/virtcontainers/pull/348

PR #429, opened 2017-08-14
docs: Update instruction to download compressed kernel

### LINKED ISSUE #414: Developer CC 3.0 install instructions points to compressed kernel only causing CC to fail
At https://github.com/clearcontainers/runtime/blob/master/docs/developers-clear-containers-install.md the instructions for the rootfs and kernel downloads a compressed kernel package only but CC 3.0 requires an uncompressed kernel for the VM. 

If you change the following variable to "demos" then the install_clear_kernel.sh script pulls from the correct repository and pulls a package with both compressed and uncompressed kernels. Either the instructions or the install script needs to be updated.

This fails
$ export kernel_clear_release=12760

This works
$ export kernel_clear_release=demos

chavafg:
Now that we have moved to use PC instead of PC-lite, we use the
vmlinuz kernel instead of vmlinux to boot a Clear Container. This
commit updates the location where the linux-container file is
downloaded, which now provides the compressed kernel.

Fixes #414.

Signed-off-by: Salvador Fuentes <salvador.fuentes@intel.com>

### DIFF
--- a/docs/developers-clear-containers-install.md
+++ b/docs/developers-clear-containers-install.md
@@ -76,36 +76,24 @@ For more details on the runtime's build system, run:
 $ make help
 ```
 
-4. Qemu-Lite
+4. Qemu-Lite, Clear Containers image and kernel
 
 In Fedora:
 ```bash
 $ source /etc/os-release
 $ sudo -E VERSION_ID=$VERSION_ID dnf config-manager --add-repo \
 http://download.opensuse.org/repositories/home:/clearcontainers:/clear-containers-3/Fedora\_$VERSION_ID/home:clearcontainers:clear-containers-3.repo
-$ sudo -E dnf install -y qemu-lite
+$ sudo -E dnf install -y qemu-lite clear-containers-image linux-container
 ```
 
-In Ubuntu:
+In Ubuntu 16.04 or newer:
 ```bash
 $ sudo -E apt-get install -y apt-transport-https ca-certificates curl software-properties-common
 $ sudo sh -c "echo 'deb http://download.opensuse.org/repositories/home:/clearcontainers:/clear-containers-3/xUbuntu_$(lsb_release -rs)/ /' >> /etc/apt/sources.list.d/clear-containers.list"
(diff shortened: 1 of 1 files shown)


docs/developers-clear-containers-install.md
```diff
@@ -102,7 +102,7 @@ $ sudo -E apt-get install -y qemu-lite
 ```bash
 $ export clear_release=$(curl -sL https://download.clearlinux.org/latest)
 $ export cc_img_path="/usr/share/clear-containers"
-$ export kernel_clear_release=12760
+$ export kernel_clear_release=demos
```
jcvenegas: could you add a comment in the script to know why we are pulling from demos ?

jcvenegas: @chavafg  also, the demos kernel must be semaphore specific, isn't it ? for developers is ok to use a newest kernel.

chavafg: true, fixing it

chavafg:
@jcvenegas fixed, thanks.

jcvenegas:
lgtm,  @sboeuf  @devimc  any comment about let developers  docs pointing to latest kernel, but keep semaphore using the old kernel?

amshinde:
@jcvenegas Developer should be using the latest kernel provided by our packaging, the kernel in demos was just a workaround for semaphore.

chavafg:
@iphutch can you please review?

docs/developers-clear-containers-install.md
```diff
@@ -76,38 +76,26 @@ For more details on the runtime's build system, run:
 $ make help
 ```
 
-4. Qemu-Lite
+4. Qemu-Lite, clear-containers image and kernel
```
iphutch: Because this isn't literal text, we should be consistent in writing "Clear Containers" with capital letters and a space. Could we also make this title a complete sentence, an option:
Qemu-Lite for Clear Containers image and kernel

jodh-intel:
Once @iphutch's comments have been addressed:

lgtm

docs/developers-clear-containers-install.md
chavafg: @iphutch fixed, as these are 3 different components that we are installing, I kept the comma. Please let me know if that is ok.

PR #440, opened 2017-08-21
cc-check: warn if unrestricted_guest unavailable under hypervisor

### LINKED ISSUE #439: cc-check should only warn if "unrestricted_guest" is not set when running under a hypervisor
Now that virtcontainers provides a way to detect if running under a hypervisor...

- https://github.com/containers/virtcontainers/pull/343

... we need to update `cc-check` to determine more intelligently if the environment is capable of running a Clear Container since although environments like Azure do not provide `unrestricted_guest`, it *is* still possible to run a Clear Container on it.

In summary, `cc-check` needs to do the following:

- if running under a hypervisor, only warn if `unrestricted_guest` is not set (since it does work under Azure for example).
- If not running under a hypervisor, error if `unrestricted_guest` is not set (no change to current behaviour).

See comments on https://github.com/clearcontainers/runtime/pull/422.

jodh-intel:
If running under a hypervisor, the kvm "unrestricted_guest" option is
not required.

Fixes #439.

### DIFF
--- a/cc-check.go
+++ b/cc-check.go
@@ -15,12 +15,14 @@
 package main
 
 import (
+	"errors"
 	"fmt"
 	"os/exec"
 	"path/filepath"
 	"regexp"
 	"strings"
 
+	vc "github.com/containers/virtcontainers"
 	"github.com/urfave/cli"
 )
 
@@ -170,6 +172,11 @@ func checkCPUAttribs(cpuinfo string, attribs map[string]string) error {
 }
 
 func checkKernelModules(modules map[string]kernelModule) error {
+	onVMM, err := vc.RunningOnVMM(procCPUInfo)
+	if err != nil {
+		return err
+	}
+
 	for module, details := range modules {
 		if !haveKernelModule(module) {
 			return fmt.Errorf("kernel module %q (%s) not found", module, details.desc)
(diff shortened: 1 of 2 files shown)


jodh-intel:
Unit test coverage maintained at 100%.

jodh-intel:
More github timeouts...

/me forces the CI to rebuild...

cc-check_test.go
```diff
@@ -373,6 +376,97 @@ func TestCheckCheckKernelModules(t *testing.T) {
...
+		// XXX: force a failure on non-VMM systems
+		{filepath.Join(sysModuleDir, "kvm_intel/parameters/unrestricted_guest"), false, "N"},
+	}
+
+	vendor := "GenuineIntel"
+	flags := "vmx lm sse4_1"
+
+	err = checkKernelModules(requiredModules)
+	// no cpuInfoFile yet
+	assert.Error(t, err)
```
jodh-intel: Good idea - done.

sboeuf:
LGTM

sboeuf:
@jodh-intel travis is complaining, I guess you need to re-upload

jodh-intel:
After many attempts, Travis is now happy.

jodh-intel:
Hi @grahamwhaley - could you take a look?

cc-check.go
```diff
@@ -189,7 +196,16 @@ func checkKernelModules(modules map[string]kernelModule) error {
 			if value == expected {
 				ccLog.Infof("Kernel module %q parameter %q has correct value", details.desc, param)
 			} else {
-				return fmt.Errorf("kernel module %q parameter %q has value %q (expected %q)", details.desc, param, value, expected)
+				msg := fmt.Sprintf("kernel module %q parameter %q has value %q (expected %q)", details.desc, param, value, expected)
+
+				// this option is not required when
+				// already running under a hypervisor.
+				if param == "unrestricted_guest" && onVMM {
```
grahamwhaley: Ideally, rather than special case this one parameter, it would be nicer if the requirement were coded into the data structures - but, this is a test program, and we only have the single case to check, so not a blocker :-)

jodh-intel: I agree. This was just the simplest pragmatic solution I could think of that didn't involve re-inventing the world ;) We might need to do that though if we start growing more exceptions like this.

PR #442, opened 2017-08-21
 test: add unit test for isCgroupMounted function

### LINKED ISSUE #405: increase unit test coverage on oci.go
Currently low coverage.

See #93.

devimc:
this patch adds the unit test for isCgroupMounted function

Fixes #405

### DIFF
--- a/oci_test.go
+++ b/oci_test.go
@@ -27,6 +27,7 @@ import (
 	"github.com/containers/virtcontainers/pkg/oci"
 	"github.com/opencontainers/runc/libcontainer/utils"
 	specs "github.com/opencontainers/runtime-spec/specs-go"
+	"github.com/stretchr/testify/assert"
 )
 
 var (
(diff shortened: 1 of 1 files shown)


jodh-intel:
lgtm

jodh-intel:
Hi @devimc - `TestIsCgroupMounted` is failing under Travis. I think this is because the test assumes the usual `/sys/fs/cgroup/...` mounts will be available. But under Travis they aren't (they are under Semaphore). See for example:

- https://travis-ci.org/jamesodhunt/procenv/jobs/233252731#L2978
 
I don't think the tests should attempt to mount the cgroup fs if not already mounted so maybe you could make it call `t.Skip()` if that test detects that `memoryCgroupPath` isn't actually mounted (you could make the test call out to `mount(8)` rather than re-implementing `isCgroupMounted()`).

devimc:
@jodh-intel  Fixed, thanks

PR #466, opened 2017-08-25
tests: Stop bundle creation timing out tests

### LINKED ISSUE #461: Tests occasionally timeout locally
The local timeout is 5 seconds:

- https://github.com/clearcontainers/runtime/blob/master/.ci/go-test.sh#L7

... but is 15 seconds when run under CI:

- https://github.com/clearcontainers/tests/blob/master/Makefile#L19

The local timeouts are due to #459 from what I can see as copying the pause bundle is much quicker than running `docker export|tar...`.

Either we increase the local timeout to match the timeout in the tests repo, or maybe we should make the runtime tests slightly more intelligent so that `main_test.go:init()` creates the OCI bundle (by calling `createRootfs()` and then `makeOCIbundle()` calls a new `copyRootfs()` that copies the fixed test bundle to `bundleDir`. That will give us the same performance characteristics as the old code (pre-#459) but will guarantee each test gets a "fresh" (pristine) bundle.

jodh-intel:
Previously, makeOCIBundle() would create a fresh bundle every time it
was called. However, since it uses docker to generate the rootfs, that
could take a while on slow systems meaning the tests could timeout.

To overcome this issue, the tests now create a full pristine bundle at
startup and makeOCIBundle() now just copies the pristine bundle to the
location requested by the calling test.

This change also removes the requirement that the bundle directory exist
before calling makeOCIBundle().

Fixes #461.

### DIFF
--- a/main_test.go
+++ b/main_test.go
@@ -56,8 +56,11 @@ const (
 	NoopAgentType  vc.AgentType      = "noop"
 )
 
-// package variables set in TestMain
-var testDir = ""
+var (
+	// package variables set by calling TestMain()
+	testDir       = ""
+	testBundleDir = ""
+)
 
 // testingImpl is a concrete mock RVC implementation used for testing
 var testingImpl = &vcMock.VCMock{}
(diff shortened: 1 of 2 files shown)


main_test.go
```diff
@@ -328,7 +340,7 @@ func createRootfs(dir string) error {
 
 // makeOCIBundle will create an OCI bundle (including the "config.json"
```
grahamwhaley: does that func name in the comment now need fixing?

jodh-intel: I see your point, but I think it's still fair to say it's making the bundle; it's just "making" it from a copy now ;)

jodh-intel:
@grahamwhaley - ta.

main_test.go
grahamwhaley: I mean the func name itself - this func is no longer called `makeOCIBundle` - now it is `realMakeOCIBundle`, no? :-)

PR #471, opened 2017-08-25
kill: Implement full unit test coverage

### LINKED ISSUE #403: increase unit test coverage on kill.go
Currently low coverage.

See #93.

sboeuf:
Relying on vcMock framework introduced in virtcontainers library,
this commit raises the unit test coverage of kill.go file to 100%.

Fixes #403

### DIFF
--- a/main_test.go
+++ b/main_test.go
@@ -19,6 +19,7 @@ package main
 import (
 	"encoding/json"
 	"errors"
+	"flag"
 	"fmt"
 	"io/ioutil"
 	"os"
@@ -33,6 +34,7 @@ import (
 	"github.com/containers/virtcontainers/pkg/vcMock"
 	"github.com/dlespiau/covertool/pkg/cover"
 	"github.com/stretchr/testify/assert"
+	"github.com/urfave/cli"
 )
 
 const (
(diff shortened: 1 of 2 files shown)


main_test.go
```diff
@@ -440,6 +442,38 @@ func writeOCIConfigFile(spec oci.CompatOCISpec, configPath string) error {
 	return ioutil.WriteFile(configPath, bytes, testFileMode)
 }
 
+func newSingleContainerPodStatusList(podID, containerID string, podState, containerState vc.State) []vc.PodStatus {
```
devimc: this cannot be generic, this should be part of each test, hence cannot be in main_test

main_test.go
```diff
@@ -440,6 +442,38 @@ func writeOCIConfigFile(spec oci.CompatOCISpec, configPath string) error {
...
+				{
+					ID:    containerID,
+					State: containerState,
+				},
+			},
+		},
+	}
+}
+
+func execCLICommandFunc(assertHandler *assert.Assertions, cliCommand cli.Command, set *flag.FlagSet, expectedErr bool) {
```
devimc: same here

main_test.go
sboeuf: Why ?

main_test.go
sboeuf: why ?

main_test.go
devimc: different tests different needs

https://github.com/clearcontainers/runtime/blob/master/list_test.go#L607
https://github.com/clearcontainers/runtime/blob/master/run_test.go#L255

sboeuf: Ok but the name of the function is specific enough so that it won't collide with other functions, right ? And I need it because I am reusing it across several tests.
We can move this later if we __really__ run into an issue but I cannot see any issue right now.

devimc:
lgtm

PR #473, opened 2017-08-28
Unit tests for exec.go

amshinde:


jodh-intel:
I've tried restarting the semaphore build a few times, but it is consistently failing.

amshinde:
@jodh-intel thanks, I'll take a look at it today.

main_test.go
```diff
@@ -457,6 +457,22 @@ func newSingleContainerPodStatusList(podID, containerID string, podState, contai
 	}
 }
 
+func newContainerListWithAnnotations(podID, containerID string, podState, containerState vc.State, annotations map[string]string) []vc.PodStatus {
```
sboeuf: You should not define a new function. Instead you should improve `newSingleContainerPodStatusList()` so that you can provide a map of annotations.

amshinde: I think its ok to have another function, we shouldnt be passing annotations when there is no need for it.

sboeuf: We don't have to pass any annotations if we don't want to, we can just give an empty/nil map.

jodh-intel: I agree with @sboeuf - this is identical to `newSingleContainerPodStatusList()` except for the inclusion of the `Annotations`.

grahamwhaley: Just a thought - you could always have `newSingleContainerPodStatusList()` call down to `newSingleContainerPodStatusListWithAnnotations()`, passing in the null annotation args etc. - so, have the no-annotations func as a thin shim/wrapper around the fuller func.

devimc:
lgtm

[Approved with PullApprove](https://pullapprove.com/clearcontainers/runtime/pull-request/473/?utm_source=github-pr&utm_medium=comment-badge&utm_campaign=clearcontainers/runtime) [Approved with PullApprove](https://pullapprove.com/clearcontainers/runtime/pull-request/473/?utm_source=github-pr&utm_medium=comment-badge&utm_campaign=clearcontainers/runtime)

main_test.go
amshinde: @sboeuf I have removed the new function and added annotations parameter to the existing function.

sboeuf:
@amshinde great thanks !
LGTM

PR #476, opened 2017-08-29
Make run tests use cached bundle

jodh-intel:


main_test.go
```diff
@@ -340,8 +340,11 @@ func createRootfs(dir string) error {
 	return nil
 }
 
-// makeOCIBundle will create an OCI bundle (including the "config.json"
+// realMakeOCIBundle will create an OCI bundle (including the "config.json"
 // config file) in the directory specified (which must already exist).
+//
+// XXX: Note that tests should *NOT* call this function - they should
+// XXX: instead call makeOCIBundle().
```
devimc: d'oh!  https://github.com/clearcontainers/runtime/blob/master/run_test.go#L194
why?

devimc:
lgtm

main_test.go
jodh-intel: There's a comment on `makeOCIBundle()`. The problem was fixed on #466.

PR #497, opened 2017-08-31
tests: Only re-run unit-tests if not initially running as root

### LINKED ISSUE #496: Only re-run unit-tests if not initially running as root
Improve `go-test.sh` so that the unit-tests are only run once if initially running as `root`. This is required since it is not clear which user `root` could switch to to run the tests as a non-priv user.

(The intent is for the tests to be run as a non-`root` user normally and that user will then elevates privs to `root` via `sudo` to re-run the tests).

jodh-intel:
Don't run the unit-tests twice if the user that invokes the tests is
already root since both runs would be running as `root`.

Also, ensure the HTML report generated by test_html_coverage() is
readable by all users.

Fixes #496.

### DIFF
--- a/.ci/go-test.sh
+++ b/.ci/go-test.sh
@@ -5,8 +5,10 @@ root_dir=`dirname $script_dir`
 
 test_packages="."
 go_test_flags="-v -race -timeout 5s"
+cov_file="profile.cov"
+tmp_cov_file="profile_tmp.cov"
 
-# Run a command as either root or non-root (current user).
+# Run a command as either root or the current user (which might still be root).
 #
 # If the first argument is "root", run using sudo, else run as normal.
 # All arguments after the first will be treated as the command to run.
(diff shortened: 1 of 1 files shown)


.ci/go-test.sh
```diff
@@ -27,24 +29,36 @@ function run_as_user
...
 
 	echo "mode: atomic" > "$cov_file"
 
 	for pkg in $test_packages; do
 
-		# Run the unit-tests *twice* (since some must run as root and
-		# others must run as non-root), combining the resulting test
-		# coverage files.
-		for user in non-root root; do
+		if [ $(id -u) -eq 0 ]
```
devimc: I think you should move this ```if ... else ``` condition before ```for pkg in $test_packages; do```

jodh-intel: good catch! moving...

jodh-intel: Branch updated.

devimc:
lgtm

Issue #463, opened 2017-08-24
create an installation doc for RHEL 7


gorozco1:
as stated in PR #372 we need to separate CentOS from RHEL and two documents need to be created.

jodh-intel:
We're now thinking that #371 can accommodate both CentOS and RHEL (it originally did, then it became CentOS only, now it's back to both :-), so this can probably be closed.

@gorozco1 - what's your view?

gorozco1:
@jodh-intel Due the fact that on RHEL system we will install `docker-ce` and an extra dependency from a `CentOS` repo the doc for RHEL should be different and warn the user to not run in prod. 
This is the reason I open #527 could you help me to review?

jodh-intel:
Hi @chavafg, @GabyCT - Is this something you could take a look into again? It's still a really important requirement, particularly so given that we have now landed https://github.com/clearcontainers/runtime/blob/master/docs/sles-installation-guide.md :smile:

/cc @xsgordon

xsgordon:
@gorozco1 Which dependency specifically has to come from CentOS? When I was attempting to follow the existing CentOS guide/script and what was pulled from where it seemed like most of the requirements were actually available in Extras/Optional for RHEL?

From my POV I'm ultimately trying to get to something aligned with the standalone CRI-O docs here:

https://github.com/kubernetes-incubator/cri-o

jodh-intel:
For reference, here's my latest "CentOS + RHEL" install script (which was eventually abandoned as we wanted to keep CentOS and RHEL separate due to their differences):

- https://github.com/jodh-intel/runtime/tree/add-centos%2Brhel-install-doc/

Specifically, the following chunk shows that for a successful install, RHEL users need to enable an extra repo:

- https://github.com/jodh-intel/runtime/blob/add-centos%2Brhel-install-doc/installation/centos%2Brhel-setup.sh#L71..L93

GabyCT:
@jodh-intel , I tried the latest script in RHEL 7.3, however, I am getting this error
install -d -m 0755 "/usr/local//bin"
install -c -m 0755 qemu-system-x86_64  "/usr/local//bin"
+ popd
~/go/src/github.com/clearcontainers/runtime/dependencies ~/go/src/github.com/clearcontainers/runtime
+ popd
~/go/src/github.com/clearcontainers/runtime
+ '[' rhel = rhel ']'
[...]

PR #372, opened 2017-07-28
docs: Create CentOS installation guide

jodh-intel:
Write a document explaining how to install Clear Containers 3.0 on a
CentOS 7 system. Added associated scripts and configuration
files (which are lightly modified versions from
https://github.com/01org/cc-oci-runtime).

Fixes #371.

Signed-off-by: James O. D. Hunt <james.o.hunt@intel.com>

PR #527, opened 2017-09-07
rhel: Installation of CC 3.0 on RHEL systems

gorozco1:
This commit includes a script based in centos installs scripts to
install clear containers 3.0 in RHEL systems.

This guide will install Docker CE from CentOS repos and extra
requeriment from CentOS repositories.

This document is not intended to be run in production systems.

Fixes: #463
Depends on: #372

Signed-off-by: Geronimo Orozco <geronimo.orozco@intel.com>

### DIFF
--- a/docs/rhel-installation-guide.md
+++ b/docs/rhel-installation-guide.md
@@ -1,0 +1,78 @@
+# Installing Intel® Clear Containers 3.0 on RHEL version 7
+
+## Scope
+
+This document covers installing Intel® Clear Containers on a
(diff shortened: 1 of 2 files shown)


jodh-intel:
@gorozco1 - one further thing to update is the final paragraph in`versions.txt` which (on #372) is documented to be for CentOS only. Once #372 lands, this PR should probably also mention RHEL to avoid any confusion.

gorozco1:
Based on discussion from dev meeting today this is not a valid scenario to install cc in rhel we need to rework and fine tuning with `docker-ce`

PR #531, opened 2017-09-08
Rename cli commands

jodh-intel:


jodh-intel:
Can we get a review of this from an approve please?

Issue #392, opened 2017-08-10
Create an architecture document


jodh-intel:
For parity with:

- https://github.com/01org/cc-oci-runtime/blob/master/documentation/architecture.md

jodh-intel:
Hi @grahamwhaley - do you wanna take this one? ;)

grahamwhaley:
heh, did I see that coming...? :-)  yes, that probably makes sense for me to take. I'll assign myself.... and everybody expect me to ask you annoying questions :-)

jodh-intel:
Ah - I'd already opened #268 for this! ;-) Closing that one...

jodh-intel:
Potentially related: https://github.com/clearcontainers/proxy/issues/104.

egernst:
@grahamwhaley I wanted to check in on this -- I would like to contribute here if it helps.  Let me know if you have any existing .mds started and/or if there are areas I can grab.

grahamwhaley:
Hi @egernst  I was going to start on this today.
My basic plan is to assess https://github.com/01org/cc-oci-runtime/blob/master/documentation/architecture.md and adopt/adapt as necessary.
I was thinking though - much of the architecture is really over in https://github.com/containers/virtcontainers, not in the runtime, as the runtime is a relatively thin shim. So, much of the documentation (detail at least) might be better placed over in https://github.com/containers/virtcontainers

Thus, I may tackle this slightly back to front - do the runtime arch docs that will reference virtcontainers, and then we can do a detailed virtcontainers doc afterwards.

grahamwhaley:
@egernst fyi, I have done a first pass clone/clean of the cc-oci-runtime docs, which can be found on my branch here as WIP:
https://github.com/grahamwhaley/runtime/tree/20170823_arch_docs/docs/architecture

There are a couple of **fixme** items (clean up the hyperstart and qemu sections, ensure we have the correct networking ref/links etc.).

And then there are some more things I need or would like to do at least:
- fix any out of date diagram bits
- probably move the diagrams to SVG - we discussed that once before. I'll see how it goes (although I am normally a little frustrated at SVG tooling :-) )
- add some sort of 'design goal' and/or overview section at the top - it is a little 'dive right in' right now for my liking.
- I think we are missing a 'storage' section, so would like to add.

[...]

PR #543, opened 2017-09-12
rfc 3.0 architecture documentation

egernst:
This is very much work in progress, and shouldn't be merged in this state.  I'm pushing this as a PR so folks can easily get their eyes on the doc/pictures and start to make suggestions edits.  This is based on Graham's initial branch with some quick cleanup on my end.  Note, the PNGs are exports from our (sorry, internal at the moment) google doc.  Once we have settled on images we can export the odp and put on to the repo as well...

Of note, I'd like to see someone write the section on the agent so it is of equivalent or more detail compared to what we had for hyperstart in 2.1.  Not sure, but perhaps @amshinde would be well suited for this?  Thoughts?

There are a few fixmes as well, which I figure will be easier to discuss during this RFC PR's review process.

Once we have input from folks, I'd recommend doing a giant squash and adding signed offs from the contributors.

Fixes https://github.com/clearcontainers/runtime/issues/392


docs/architecture/architecture.md
```diff
@@ -0,0 +1,378 @@
...
+** fixme - discuss different QEMUs - lite, q35, pc etc. **
+Although Clear Containers can run with any recent QEMU release, containers boot time and memory
+footprint are significantly optimized by using a specific QEMU version called [`qemu-lite`](https://github.com/clearcontainers/qemu/tree/qemu-lite-v2.9.0).
+
+`qemu-lite` improvements comes through a new `pc-lite` machine type, mostly by:
+- Removing many of the legacy hardware devices support so that the guest kernel does not waste
+time initializing devices of no use for containers.
+- Skipping the guest BIOS/firmware and jumping straight to the Clear Containers kernel.
+
+#### Agent
```
egernst: @amshinde ; can you help with this section?

docs/architecture/architecture.md
```diff
@@ -0,0 +1,378 @@
...
+management commands to the agent while the I/O serial device is used to pass I/O streams (`stdout`,
+`stderr`, `stdin`) between the guest and the Docker Engine.
+
+For any given container, both the init process and all potentially executed commands within that
+container, together with their related I/O streams, need to go through 2 virtio serial interfaces
+exported by QEMU. The [Clear Containers proxy (`cc-proxy`)](https://github.com/clearcontainers/proxy)
+multiplexes and demultiplexes those commands and streams for all container virtual machines.
+There is only one `cc-proxy` instance running per Clear Containers host.
+
+On the host, each container process is reaped by a Docker specific (`containerd-shim`) monitoring
```
egernst: This is pretty docker specific.  We may want to make this more generic so it can cover !docker as well.

docs/architecture/architecture.md
```diff
@@ -0,0 +1,378 @@
...
+can not monitor, control or reap them. `cc-runtime` fixes that issue by creating an
+[additional shim process (`cc-shim`)](https://github.com/clearcontainers/shim)
+between `containerd-shim` and `cc-proxy`. A `cc-shim` instance will both forward signals and `stdin`
+streams to the container process on the guest and pass the container `stdout` and `stderr` streams
+back to the Docker engine via `containerd-shim`.
+`cc-runtime` creates a `cc-shim` daemon for each Docker container and for each command Docker
+wants to run within an already running container (`docker exec`).
+
+The container workload, i.e. the actual OCI bundle rootfs, is exported from the host to
+the virtual machine via a 9pfs virtio mount point. Hyperstart uses this mount point as the root
```
egernst: s/Hyperstart/agent

docs/architecture/architecture.md
```diff
@@ -0,0 +1,378 @@
...
+Clear Containers use [KVM](http://www.linux-kvm.org/page/Main_Page)/[QEMU](http://www.qemu-project.org/) to
+create virtual machines where Docker containers will run:
+
+![QEMU/KVM](qemu.png)
+
+** fixme - discuss different QEMUs - lite, q35, pc etc. **
+Although Clear Containers can run with any recent QEMU release, containers boot time and memory
+footprint are significantly optimized by using a specific QEMU version called [`qemu-lite`](https://github.com/clearcontainers/qemu/tree/qemu-lite-v2.9.0).
+
+`qemu-lite` improvements comes through a new `pc-lite` machine type, mostly by:
```
egernst: We are using PC machine type for 3.0, I think for release perhaps it makes sense to just remove all of the permutations and just describe what we actually have?

docs/architecture/architecture.md
```diff
@@ -0,0 +1,378 @@
...
+- Removing many of the legacy hardware devices support so that the guest kernel does not waste
+time initializing devices of no use for containers.
+- Skipping the guest BIOS/firmware and jumping straight to the Clear Containers kernel.
```
egernst: not really the case for 3.0 release, right?

docs/architecture/architecture.md
```diff
@@ -0,0 +1,378 @@
...
+1. Create the container namespaces (Only the network and mount namespaces are currently supported),
+according to the container OCI configuration file.
+2. Spawn the `cc-shim` process and have it wait on a couple of temporary pipes for:
+   * A `cc-proxy` created file descriptor (one end of a socketpair) for the shim to connect to.
+   * The container `agent` sequence numbers for at most 2 I/O streams (One for `stdout` and `stdin`, one for `stderr`).
+   The `agent` uses those sequence numbers to multiplex all streams for all processes through one serial interface (The
+   virtio I/O serial one).
+3. Run all the [OCI hooks](https://github.com/opencontainers/runtime-spec/blob/master/config.md#hooks) in the container namespaces,
+as described by the OCI container configuration file.
+4. **fixme** [Set up the container networking](https://github.com/clearcontainers/runtime/blob/master/documentation/architecture.md#networking).
```
egernst: fixme is just for the link location, right @grahamwhaley ?

egernst:
@sameo @jodh-intel @mcastelino - if you have a chance to start taking a look.  @jcvenegas perhaps we can reference the proxy protocol in the proxy section?  

@iphutch  -- early heads up and FYI that this is (in early stage of) the pipeline

docs/architecture/architecture.md
```diff
@@ -0,0 +1,378 @@
...
+
+`qemu-lite` improvements comes through a new `pc-lite` machine type, mostly by:
+- Removing many of the legacy hardware devices support so that the guest kernel does not waste
+time initializing devices of no use for containers.
+- Skipping the guest BIOS/firmware and jumping straight to the Clear Containers kernel.
+
+#### Agent
+
+['agent'](https://github.com/clearcontainers/agent) is a daemon running on the guest as an
+agent for managing containers and processes potentially running within those containers.
```
jodh-intel: I'd rephrase this to avoid the "double agent". How about "as a supervisor for managing..."?

docs/architecture/architecture.md
```diff
@@ -0,0 +1,378 @@
...
+
+##### `create`
+
+When handling the OCI `create` command, `cc-runtime` goes through the following steps:
+
+1. Create the container namespaces (Only the network and mount namespaces are currently supported),
+according to the container OCI configuration file.
+2. Spawn the `cc-shim` process and have it wait on a couple of temporary pipes for:
+   * A `cc-proxy` created file descriptor (one end of a socketpair) for the shim to connect to.
+   * The container `agent` sequence numbers for at most 2 I/O streams (One for `stdout` and `stdin`, one for `stderr`).
```
jodh-intel: s/2/two/

docs/architecture/architecture.md
```diff
@@ -0,0 +1,378 @@
...
+10. The `cc-shim` instance is put into a stopped state to prevent it from doing I/O before the container is started.
+
+![Docker create](create.png)
+
+At that point, the container sandbox is created in the virtual machine and `cc-shim` is stopped on the host.
+However the container process itself is not yet running as one needs to call `docker start` to actually start it.
+
+##### `start`
+
+On namespace containers `start` launches a traditional Linux container process in its own set of namespaces.
```
jodh-intel: s/On namespace/With namespace/.

docs/architecture/architecture.md
```diff
@@ -0,0 +1,378 @@
...
+With Clear Containers, the main task of `cc-runtime` is to create and start a container within the
+pod that got created during the `create` step. In practice, this means `cc-runtime` follows
```
jodh-intel: s/that got/that was/

docs/architecture/architecture.md
```diff
@@ -0,0 +1,378 @@
...
+#### Proxy
+
+`cc-proxy` is a daemon offering access to the VM [`agent`](https://github.com/clearcontainers/agent)
+to multiple `cc-shim` and `cc-runtime` clients.
+Only a single instance of `cc-proxy` per host is necessary as it can be used for several different VMs.
+Its main role is to:
+- Arbitrate access to the `agent` control channel between all the `cc-runtime` instances and the `cc-shim` ones.
+- Route the I/O streams between the various `cc-shim` instances and the `agent`.
+
+`cc-proxy` provides 2 client interfaces:
```
jodh-intel: s/2/two/

docs/architecture/architecture.md
```diff
@@ -0,0 +1,378 @@
...
+- `Attach`: `cc-runtime` uses that command as a VM multiplexer as it allows it to notify `cc-proxy` about
+which VM it wants to talk to. In other words, this commands allows `cc-runtime` to attach itself to a
+running VM.
+- `AllocateIO`: As the `agent` can potentially handle I/O streams from multiple container processes at the
+same time, it needs to be able to associate any given stream to a container process. This is done by the `agent`
+allocating a set of at most 2 so-called sequence numbers per container process. `cc-runtime` will send
+the `AllocateIO` command to `cc-proxy` to have it request the `agent` to allocate those sequence numbers.
+They will be passed as command line arguments to `cc-shim`, who will then use them to e.g. prepend its stdin
+stream packets with the right sequence number.
+- `Hyper`: This command is used by both `cc-runtime` and `cc-shim` to forward `agent` specific
```
jodh-intel: @jcvenegas - can you review this section?

docs/architecture/architecture.md
```diff
@@ -0,0 +1,378 @@
...
+commands.
+
+For more details about `cc-proxy`'s protocol, theory of operations or debugging tips, please read
```
jodh-intel: Is it worth adding an explicit link to @jcvenegas's new proxy UML sequence diagram?

docs/architecture/architecture.md
```diff
@@ -0,0 +1,378 @@
...
+[`cc-proxy` README](https://github.com/clearcontainers/proxy).
+
+#### Shim
```
jodh-intel: @amshinde - can you review this section please?

docs/architecture/architecture.md
```diff
@@ -0,0 +1,378 @@
...
+details on how `cc-runtime` implements [CNM](https://github.com/docker/libnetwork/blob/master/docs/design.md).
+
+## Appendices
+
+### DAX
+
+Clear Containers utilises the Linux kernel DAX (Direct Access filesystem)
+feature to efficiently map some host side files into the guest VM space.
+In particular, Clear Containers uses the `QEMU` nvdimm feature to provide a
+memory mapped virtual device that can be used to DAX map the mini-OS root
```
jodh-intel: This seems to be the first mention of the mini-OS. I think it might be helpful to introduce it when we first start taking about the VM, pointing out that there is a mini-OS and the 9p-mapped docker image (`ubuntu`, `busybox`, etc).

docs/architecture/architecture.md
```diff
@@ -0,0 +1,378 @@
...
+filesystem into the guest space.
+
+Mapping files using DAX provides a number of benefits over more traditional VM
+file and device mapping mechanisms:
+
+- Mapping as a direct access devices allows the guest to directly access
+the memory pages (such as via eXicute In Place (XIP)), bypassing the guest
```
jodh-intel: s/the memory pages/the host memory pages/ ?

docs/architecture/architecture.md
```diff
@@ -0,0 +1,378 @@
...
+page cache. This provides both time and space optimisations.
+- Mapping as a direct access device inside the VM allows pages from the
+host to be demand loaded using page faults, rather than having to make requests
+via a virtualised device (causing expensive VM exits/hypercalls), thus providing
+a speed optimisation.
+- Utilising shmem MAP_SHARED on the host allows the host to efficiently
```
jodh-intel: s/shmem MAP_SHARED/`MAP_SHARED` shared memory/ ?

docs/architecture/architecture.md
```diff
@@ -0,0 +1,378 @@
...
+
+![DAX](DAX.png)
+
+More information about DAX can be found in the Linux Kernel
+[documentation](https://git.kernel.org/cgit/linux/kernel/git/torvalds/linux.git/tree/Documentation/filesystems/dax.txt)
+
+Information on the use of nvdimm via QEMU is available in the QEMU source code
+[here](http://git.qemu-project.org/?p=qemu.git;a=blob;f=docs/nvdimm.txt;hb=HEAD)
+
+### Previous releases
```
jodh-intel: Maybe this should be called "Architectural changes by release" or something? We can then document here what changed between 2.1 and 3.0.

docs/architecture/architecture.md
```diff
@@ -0,0 +1,378 @@
...
+    * [Previous Releases](#previous-releases)
+        * [V2.0](#version-20)
+        * [V1.0](#version-10)
+
+## Overview
+
+This is an architectural overview of Clear Containers, based on the 3.0 release.
+
+The [Clear Containers runtime (cc-runtime)](https://github.com/clearcontainers/runtime)
+complies with the [OCI](https://github.com/opencontainers) specifications and thus
```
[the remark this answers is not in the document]
grahamwhaley: agree - 'compatible'

docs/architecture/architecture.md
grahamwhaley: Looks like it. I believe we can change it to a 'local reference', so just `#networking` should resolve to the section in this document.

docs/architecture/architecture.md
```diff
@@ -0,0 +1,378 @@
...
+![Overall architecture](overall-architecture.png)
+
+#### Hypervisor
+
+Clear Containers use [KVM](http://www.linux-kvm.org/page/Main_Page)/[QEMU](http://www.qemu-project.org/) to
+create virtual machines where Docker containers will run:
+
+![QEMU/KVM](qemu.png)
+
+** fixme - discuss different QEMUs - lite, q35, pc etc. **
```
grahamwhaley: I think for this release we can drop this 'fixme' and just mention qemu-lite.

docs/architecture/architecture.md
```diff
@@ -0,0 +1,378 @@
...
+   The `agent` uses those sequence numbers to multiplex all streams for all processes through one serial interface (The
+   virtio I/O serial one).
+3. Run all the [OCI hooks](https://github.com/opencontainers/runtime-spec/blob/master/config.md#hooks) in the container namespaces,
+as described by the OCI container configuration file.
+4. **fixme** [Set up the container networking](https://github.com/clearcontainers/runtime/blob/master/documentation/architecture.md#networking).
+This must happen after all hooks are done as one of them is potentially setting
+the container networking namespace up.
+5. Create the virtual machine running the container process. The VM `systemd` instance will spawn the `agent` daemon.
+6. Wait for the `agent` to signal that it is ready.
+7. Send the pod creation command to the `agent`. The `agent` pod is the container process sandbox.
```
sameo: With 3.0, the pod creation command is sent at `start` time.

docs/architecture/architecture.md
```diff
@@ -0,0 +1,378 @@
...
+8. Send the allocateIO command to the proxy, for getting the `agent` I/O sequence numbers described in step 2.
+9. Pass the `cc-proxy` socketpair file descriptor, and the I/O sequence numbers to the listening cc-shim process through the dedicated pipes.
+10. The `cc-shim` instance is put into a stopped state to prevent it from doing I/O before the container is started.
```
sameo: 8 and 9 can be removed with 3.0
10 is not accurate. The shim is alive after a create.

docs/architecture/architecture.md
```diff
@@ -0,0 +1,378 @@
...
+##### `start`
+
+On namespace containers `start` launches a traditional Linux container process in its own set of namespaces.
+With Clear Containers, the main task of `cc-runtime` is to create and start a container within the
+pod that got created during the `create` step. In practice, this means `cc-runtime` follows
+these steps:
+
+1. `cc-runtime` connects to `cc-proxy` and sends it the `attach` command to let it know which pod
+we want to use to create and start the new container.
+2. `cc-runtime` sends an agent `NEWCONTAINER` command to create and start a new container in
```
sameo: Before that step, we also send the `STARTPOD` command, and then one `NEWCONTAINER` per container on the pod.

docs/architecture/architecture.md
```diff
@@ -0,0 +1,378 @@
...
+All I/O streams from the executed command will be passed back to Docker through a newly
+created `cc-shim`.
+
+The `exec` code path is partly similar to the `create` one and `cc-runtime` goes through
+the following steps:
+
+1. `cc-runtime` connects to `cc-proxy` and sends it the `attach` command to let it know which pod
+we want to use to run the `exec` command.
+2. `cc-runtime` sends the allocateIO command to the proxy, for getting the `agent` I/O sequence
+numbers for the `exec` command I/O streams.
```
sameo: This has also changed with 3.0. We connect to the proxy and get a token back. We then create a shim with the given token and the shim connects to the proxy.

docs/architecture/architecture.md
```diff
@@ -0,0 +1,378 @@
...
+delete all related proxy resources:
+
+1. `cc-runtime` connects to `cc-proxy` and sends it the `attach` command to let it know
+on which pod the container it is trying to to `delete` is running.
+2. `cc-runtime` sends an agent `DESTROYPOD` command to `destroy` the pod holding the
+container running on the guest. The command is sent to `cc-proxy` who forwards it to the right
+agent instance running in the appropriate guest.
+3. After deleting the last running pod, the `agent` will gracefully shut the virtual machine
+down.
+4. `cc-runtime` sends the `BYE` command to `cc-proxy`, to let it know that a given virtual
```
sameo: `BYE` has been replaced with `UnregisterVM`

docs/architecture/architecture.md
```diff
@@ -0,0 +1,378 @@
...
+Only a single instance of `cc-proxy` per host is necessary as it can be used for several different VMs.
+Its main role is to:
+- Arbitrate access to the `agent` control channel between all the `cc-runtime` instances and the `cc-shim` ones.
+- Route the I/O streams between the various `cc-shim` instances and the `agent`.
+
+`cc-proxy` provides 2 client interfaces:
+
+- A UNIX, named socket for all `cc-runtime` instances on the host to send commands to `cc-proxy`.
+- One socket pair per `cc-shim` instance, to send stdin and receive stdout and stderr I/O streams. See the
+[cc-shim section](#shim)
```
sameo: That has changed as well and is now simplified. There is one single socket (UNIX or TCP) for all runtimes and all shims.

docs/architecture/architecture.md
sameo: This is well out of date as well for 3.0.

docs/architecture/architecture.md
sameo: We should definitely have a link to it.

docs/architecture/architecture.md
```diff
@@ -0,0 +1,378 @@
...
+        * [V1.0](#version-10)
+
+## Overview
+
+This is an architectural overview of Clear Containers, based on the 3.0 release.
+
+The [Clear Containers runtime (cc-runtime)](https://github.com/clearcontainers/runtime)
+complies with the [OCI](https://github.com/opencontainers) specifications and thus
+works seamlessly with the [Docker Engine](https://www.docker.com/products/docker-engine)
+pluggable runtime architecture. In other words, one can transparently replace the
```
mcastelino: I would not say replace. We can add another runtime. We cannot satisfy all possible types of containers like priv ones.

docs/architecture/architecture.md
```diff
@@ -0,0 +1,378 @@
...
+![Runtime and virtcontainers](runtime-vc-relationship.png)
+![Docker and Clear Containers](docker-cc.png)
+
+`cc-runtime` creates a QEMU/KVM virtual machine for each container the Docker engine creates.
+
+The container process is then spawned by an [agent](https://github.com/clearcontainers/agent),
+running as a daemon on the guest operating system.
+Hyperstart opens 2 virtio serial interfaces (Control and I/O) on the guest, and QEMU exposes them
+as serial devices on the host. `cc-runtime` uses the control device for sending container
+management commands to the agent while the I/O serial device is used to pass I/O streams (`stdout`,
```
mcastelino: Would be good to point to a section that contain the container management command or maybe the source file where the command protocol is defined.

docs/architecture/architecture.md
```diff
@@ -0,0 +1,378 @@
...
+`stderr`, `stdin`) between the guest and the Docker Engine.
+
+For any given container, both the init process and all potentially executed commands within that
```
mcastelino: Would be good to document how exec is handled too.

docs/architecture/architecture.md
```diff
@@ -0,0 +1,378 @@
...
+- Skipping the guest BIOS/firmware and jumping straight to the Clear Containers kernel.
+
+#### Agent
+
+['agent'](https://github.com/clearcontainers/agent) is a daemon running on the guest as an
+agent for managing containers and processes potentially running within those containers.
+
+The `agent` execution unit is the pod. An `agent` pod is a container sandbox defined
+by a set of namespaces (UTS, PID, mount and IPC). Although a pod can hold several containers,
+`cc-runtime` always runs a single container per pod. **fixme** incorrect<--
```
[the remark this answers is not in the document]
mcastelino: This is not accurate
> `cc-runtime` always runs a single container per pod.

docs/architecture/architecture.md
```diff
@@ -0,0 +1,378 @@
...
+
+#### Runtime
+
+`cc-runtime` is an OCI compatible container runtime and is responsible for handling all
+commands specified by [the OCI runtime specification](https://github.com/opencontainers/runtime-spec)
+and launching `cc-shim` instances.
+
+Here we will describe how `cc-runtime` handles the most important OCI commands.
+
+##### `create`
```
mcastelino: Would be nice to link to the source code file that implements create

docs/architecture/architecture.md
amshinde: @egernst I'll take a look.

docs/architecture/architecture.md
egernst: done

egernst:
Moved this document to https://github.com/clearcontainers/runtime/wiki/Clear-Containers-Architecture to facilitate better collaboration at this point in the review process.

docs/architecture/architecture.md
egernst: good idea @jodh-intel .  I made this update on the wiki: https://github.com/clearcontainers/runtime/wiki/Clear-Containers-Architecture#architectural-changes-by-release

docs/architecture/architecture.md
egernst: done.

docs/architecture/architecture.md
egernst: done.

docs/architecture/architecture.md
egernst: done

docs/architecture/architecture.md
egernst: done

docs/architecture/architecture.md
egernst: done.

docs/architecture/architecture.md
egernst: done

docs/architecture/architecture.md
egernst: done

docs/architecture/architecture.md
egernst: took a pass at this @ https://github.com/clearcontainers/runtime/wiki/Clear-Containers-Architecture#hypervisor - not sure how much detail we should go into features we aren't actually using though....

docs/architecture/architecture.md
egernst: ack

docs/architecture/architecture.md
```diff
@@ -0,0 +1,378 @@
...
+
+The container workload, i.e. the actual OCI bundle rootfs, is exported from the host to
+the virtual machine via a 9pfs virtio mount point. Hyperstart uses this mount point as the root
+filesystem for the container processes.
+
+![Overall architecture](overall-architecture.png)
+
+#### Hypervisor
+
+Clear Containers use [KVM](http://www.linux-kvm.org/page/Main_Page)/[QEMU](http://www.qemu-project.org/) to
```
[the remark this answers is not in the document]
egernst: done

docs/architecture/architecture.md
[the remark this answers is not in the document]
egernst: done

egernst: ack

docs/architecture/architecture.md
egernst: done

docs/architecture/architecture.md
```diff
@@ -0,0 +1,378 @@
...
+
+The container process is then spawned by an [agent](https://github.com/clearcontainers/agent),
+running as a daemon on the guest operating system.
+Hyperstart opens 2 virtio serial interfaces (Control and I/O) on the guest, and QEMU exposes them
+as serial devices on the host. `cc-runtime` uses the control device for sending container
+management commands to the agent while the I/O serial device is used to pass I/O streams (`stdout`,
+`stderr`, `stdin`) between the guest and the Docker Engine.
+
+For any given container, both the init process and all potentially executed commands within that
+container, together with their related I/O streams, need to go through 2 virtio serial interfaces
```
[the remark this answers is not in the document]
egernst: done

docs/architecture/architecture.md
```diff
@@ -0,0 +1,378 @@
...
+[default Docker runtime (runc)](https://github.com/opencontainers/runc) with `cc-runtime`.
+
+![Runtime and virtcontainers](runtime-vc-relationship.png)
+![Docker and Clear Containers](docker-cc.png)
+
+`cc-runtime` creates a QEMU/KVM virtual machine for each container the Docker engine creates.
+
+The container process is then spawned by an [agent](https://github.com/clearcontainers/agent),
+running as a daemon on the guest operating system.
+Hyperstart opens 2 virtio serial interfaces (Control and I/O) on the guest, and QEMU exposes them
```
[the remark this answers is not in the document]
egernst: ack

docs/architecture/architecture.md
```diff
@@ -0,0 +1,378 @@
...
+pluggable runtime architecture. In other words, one can transparently replace the
+[default Docker runtime (runc)](https://github.com/opencontainers/runc) with `cc-runtime`.
+
+![Runtime and virtcontainers](runtime-vc-relationship.png)
+![Docker and Clear Containers](docker-cc.png)
+
+`cc-runtime` creates a QEMU/KVM virtual machine for each container the Docker engine creates.
+
+The container process is then spawned by an [agent](https://github.com/clearcontainers/agent),
+running as a daemon on the guest operating system.
```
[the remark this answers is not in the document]
egernst: done

docs/architecture/architecture.md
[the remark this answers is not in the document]
egernst: done

docs/architecture/architecture.md
jcvenegas: Updated to 3.0 proxy protocol in wiki

docs/architecture/architecture.md
jcvenegas: Added link to the UML sequence diagram

docs/architecture/architecture.md
jcvenegas: Added url to to godoc proxy api protocol, and added link to UML sequence diagram.

docs/architecture/architecture.md
```diff
@@ -0,0 +1,378 @@
...
+
+When handling the OCI `create` command, `cc-runtime` goes through the following steps:
+
+1. Create the container namespaces (Only the network and mount namespaces are currently supported),
+according to the container OCI configuration file.
+2. Spawn the `cc-shim` process and have it wait on a couple of temporary pipes for:
+   * A `cc-proxy` created file descriptor (one end of a socketpair) for the shim to connect to.
+   * The container `agent` sequence numbers for at most 2 I/O streams (One for `stdout` and `stdin`, one for `stderr`).
+   The `agent` uses those sequence numbers to multiplex all streams for all processes through one serial interface (The
+   virtio I/O serial one).
```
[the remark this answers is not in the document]
egernst: I'll start taking a first pass on updating this section unless I hear from you @sameo .

I wonder if I should be detailing virtcontainers here, then...

docs/architecture/architecture.md
egernst: @jcvenegas -- can you take a look and update?

jcvenegas: @egernst  I just check it is updated describing  the interaction between proxy and shim in the new protocol

egernst:
@iphutch  - Can you start taking a look at this PR?

docs/architecture/architecture.md
```diff
@@ -0,0 +1,626 @@
...
+container process. As the container process reaper runs on the host, it cannot directly monitor a process running
+within a virtual machine. At most it can see the QEMU process, but that is not enough.
+With Clear Containers, `cc-shim` acts as the container process that the container process reaper can monitor. Therefore
+`cc-shim` needs to handle all container I/O streams (`stdout`, `stdin` and `stderr`) and forward all signals the
+container process reaper decides to send to the container process.
+
+`cc-shim` has an implicit knowledge about which VM agent will handle those streams and signals and thus acts as
+an encapsulation layer between the container process reaper and the `cc-agent`. `cc-shim`:
+
+- Connects to `cc-proxy` using a token obtained by calling the `cc-proxy` `ConnectShim` command. The token is passed from `cc-runtime` to `cc-shim` when the former spawns the latter and is used to identify the true container process that the shim process will be shadowing or representing.
```
iphutch: line length & indentation for this and the Proxy section

docs/architecture/architecture.md
```diff
@@ -0,0 +1,626 @@
...
+![CNI Diagram](arch-images/CNI_diagram.png)
+
+__Runtime network setup with CNI__
+
+1. Create the network namespace
+
+2. Get CNI plugin information
+
+3. Start the plugin (providing previously created netns) to add a network
+described into /etc/cni/net.d/ directory. At that time, the CNI plugin will
```
iphutch: use literal text for the directory name

docs/architecture/architecture.md
```diff
@@ -0,0 +1,626 @@
...
+create the cni0 network interface and a veth pair between the host and the created
+netns. It links cni0 to the veth pair before to exit.
+
+4. Create bridge, TAP, and link all together with network interface previously
+created
+
+5. Start VM inside the netns and start the container
+
+## Storage
+Container workloads are shared with the virtualized environment through 9pfs.
```
iphutch: can we link to information on 9pfs?

docs/architecture/architecture.md
```diff
@@ -0,0 +1,626 @@
...
+
+Users can check to see if the container uses devicemapper block device as its
+rootfs by calling `mount(8)` within the counter.  If devicemapper block device
+is used, '/' will be mounted on `/dev/vda`.
+
+# Appendices
+
+## DAX
+
+Clear Containers utilises the Linux kernel DAX (Direct Access filesystem)
```
iphutch: Ditto, can we link to DAX info here?:
[Direct Access filesystem](https://git.kernel.org/cgit/linux/kernel/git/torvalds/linux.git/tree/Documentation/filesystems/dax.txt)

docs/architecture/architecture.md
```diff
@@ -0,0 +1,626 @@
...
+page cache. This provides both time and space optimisations.
+- Mapping as a direct access device inside the VM allows pages from the
+host to be demand loaded using page faults, rather than having to make requests
+via a virtualised device (causing expensive VM exits/hypercalls), thus providing
+a speed optimisation.
+- Utilising MAP_SHARED shared memory on the host allows the host to efficiently
+share pages.
+
+Clear Containers uses the following steps to set up the DAX mappings:
+- QEMU is configured with an nvdimm memory device, with a memory file
```
iphutch: Use numbers instead of - here. Also, indentation

docs/architecture/architecture.md
```diff
@@ -0,0 +1,626 @@
...
+- The guest kernel command line mounts this nvdimm device with the DAX
+feature enabled, allowing direct page mapping and access, thus bypassing the
+guest page cache.
+
+![DAX](arch-images/DAX.png)
+
+More information about DAX can be found in the Linux Kernel
+[documentation](https://git.kernel.org/cgit/linux/kernel/git/torvalds/linux.git/tree/Documentation/filesystems/dax.txt)
+
+Information on the use of nvdimm via QEMU is available in the QEMU source code
```
iphutch: make "QEMU source code" the link and remove "here"

egernst:
@iphutch -- Thanks!  Will start to address these now...

docs/architecture/architecture.md
```diff
@@ -0,0 +1,626 @@
...
+`cc-shim` needs to handle all container I/O streams (`stdout`, `stdin` and `stderr`) and forward all signals the
+container process reaper decides to send to the container process.
+
+`cc-shim` has an implicit knowledge about which VM agent will handle those streams and signals and thus acts as
+an encapsulation layer between the container process reaper and the `cc-agent`. `cc-shim`:
+
+- Connects to `cc-proxy` using a token obtained by calling the `cc-proxy` `ConnectShim` command. The token is passed from `cc-runtime` to `cc-shim` when the former spawns the latter and is used to identify the true container process that the shim process will be shadowing or representing.
+- Fragments and encapsulates the standard input stream from the container process reaper into `cc-proxy` stream frames:
+```
+                     1 1 1 1 1 1 1 1 1 2 2 2 2 2 2 2 2 2 2 3 3
```
egernst: @amshinde  -- what's the intention for the above line?

amshinde: good question , I had asked the same to damien with no answer on that .
This is a direct copy of the frame format from his documented proxy protocol :)

docs/architecture/architecture.md
```diff
@@ -0,0 +1,683 @@
...
+## Previous Releases
+
+This section provides a brief overview of architectural features enabled in
+prior and current versions of Clear Containers.
+
+### Version 3.0
+
+- Moved from hyperstart to `cc-agent` as an agent inside the VM.
+- Moved from `qemu-lite` to `pc` QEMU machine type.
+- Rewrite of runtime in go, leveraging virtcontainers.
```
sameo: I'd add a few items here:

- virtio-blk for block based graph drivers
- New simplified protocol between the shim, proxy and runtime
- KSM throttling

docs/architecture/architecture.md
```diff
@@ -0,0 +1,677 @@
...
+__Runtime network setup with CNM__
+
+1. Read config.json
+
+2. Create the network namespace
+
+3. Call the prestart hook (from inside the netns)
+
+4. Scan network interfaces inside netns and get the name of the interface
+created by prestart hook
```
iphutch: Indentation :( From here down. line 525, 538, 543, 592, 595,599, 651, and finally (fittingly) 666.

docs/architecture/architecture.md
```diff
@@ -0,0 +1,677 @@
...
+about completely deleting the whole thing: the pod, its containers, all `cc-shim`
+instances and the virtual machine itself.
+Also, the latter case is only supported by Kubernetes whereas Docker always asks
+for the former one.
+
+Below we will focus on describing the entire pod deletion process and we will
+look at the generic, multi-containers pod setup. Docker is just a particular case
+for single container pods.
+
+1. `cc-runtime` connects to `cc-proxy` and sends it the `attach` command to let
```
iphutch: Indentation in this section

docs/architecture/architecture.md
```diff
@@ -0,0 +1,677 @@
...
+PID, mount and IPC namespaces with the container's init process.
+
+#### [`kill`](https://github.com/clearcontainers/runtime/blob/master/kill.go)
+
+When sending the OCI `kill` command, container runtimes should send a [UNIX signal](https://en.wikipedia.org/wiki/Unix_signal)
+to the container process.
+In the Clear Containers context, this means `cc-runtime` needs to handle 2 resources:
+
+1. It needs to send the signal to the container running on the VM.
+2. It needs to signal the [`shim`](#shim) since we don't want to leave dangling
```
iphutch: indentation, also in 316's list

docs/architecture/architecture.md
```diff
@@ -0,0 +1,674 @@
...
+is compatible with the [OCI](https://github.com/opencontainers) [runtime specification](https://github.com/opencontainers/runtime-spec)
+and therefore works seamlessly with the
+[Docker\* Engine](https://www.docker.com/products/docker-engine) pluggable runtime
+architecture. It also supports the [Kubernetes\* Container Runtime Interface (CRI)](https://github.com/kubernetes/kubernetes/tree/master/pkg/kubelet/apis/cri/v1alpha1/runtime)
+through the [CRI-O\*](https://github.com/kubernetes-incubator/cri-o) implementation.
+In other words, you can transparently select between the
+[default Docker and CRI-O runtime (runc)](https://github.com/opencontainers/runc)
+and `cc-runtime`.
+
+![Runtime and virtcontainers](arch-images/runtime-vc-relationship.png)
```
iphutch: This .png isn't included in the PR or has a different name. It will bring a 404 error.

docs/architecture/architecture.md
```diff
@@ -0,0 +1,674 @@
...
+
+In order to provide some advanced features which were lacking in `pc-lite` like hotplug
+support, the default was changed to `pc`.
+
+In the future, Clear Containers plan to move to a 2.9 based version of QEMU,
+available at 
+the [Clear Containers QEMU repo](https://github.com/clearcontainers/qemu/tree/qemu-lite-v2.9.0).
+This transition has been delayed until after the release of Clear Containers 3.0
+due to regressions, as described in [runtime issue 407]
+(https://github.com/clearcontainers/runtime/issues/407). Once support for
```
iphutch: Breaking up this [runtime issue 407](https://github.com/clearcontainers/runtime/issues/407) breaks the link's markup. Don't worry about going over 78chars when it's code or markup forcing you over:
due to regressions, as described in [runtime issue 407](https://github.com/clearcontainers/runtime/issues/407).
Once support for

docs/architecture/architecture.md
```diff
@@ -0,0 +1,674 @@
...
+Users can check to see if the container uses devicemapper block device as its
+rootfs by calling `mount(8)` within the counter.  If devicemapper block device
+is used, '/' will be mounted on `/dev/vda`.
+
+# Appendices
+
+## DAX
+
+Clear Containers utilises the Linux kernel DAX [(Direct Access filesystem)]
+(https://git.kernel.org/cgit/linux/kernel/git/torvalds/linux.git/tree/Documentation/filesystems/dax.txt)
```
iphutch: Build error. Ignore line length rules for markup OR bring the whole link to 575.

egernst: Whoops - thanks

docs/architecture/architecture.md
egernst: I meant to fix all those -- thanks!  Done

docs/architecture/architecture.md
egernst: y'ouch - git blame says that wa sfrom me, but I don't recall adding this, nor can I find this image anywehere... removing now...

docs/architecture/architecture.md
```diff
@@ -0,0 +1,671 @@
...
+the [Clear Containers QEMU repo](https://github.com/clearcontainers/qemu/tree/qemu-lite-v2.9.0).
+This transition has been delayed until after the release of Clear Containers 3.0
+due to regressions, as described in [runtime issue 407](https://github.com/clearcontainers/runtime/issues/407).
+Once support for features like hotplug are available in `Q35`, the project will
+look to transition to this machine type.
+ 
+## Agent
+
+[`cc-agent`](https://github.com/clearcontainers/agent) is a daemon running in the
+guest as a supervisor for managing containers and processes potentially running
```
jodh-intel: The word "potentially" is redundant so I'd drop it.

jodh-intel:
You might like to add in some `Contributions-by:` in the commit like we have here: 56e96d3b788dc91bf144e629c7200de98cbc5ff3.

docs/architecture/architecture.md
grahamwhaley: agreed - was just about to say the same thing.

docs/architecture/architecture.md
```diff
@@ -0,0 +1,671 @@
...
+
+`cc-agent` uses a communication protocol defined by the [hyperstart](https://github.com/hyperhq/hyperstart)
+project. This was chosen to maintain backward compatibility with the `hyperstart`
+agent used in the 2.x `Clear Containers` architecture.
+
+The `cc-agent` interface consists of:
+- A control serial channel over which the `cc-agent` sends and receives specific
+  commands for controlling and managing pods and containers. Detailed information
+  about the commands can be found at [`cc-agent` API](https://github.com/clearcontainers/agent/tree/master/api).
+- An I/O serial channel for passing the container processes output streams (stdout,
```
grahamwhaley: earlier in the doc we backtick std[in|out|err] to `stdin` - we should consider doing that consistently through the document.

docs/architecture/architecture.md
```diff
@@ -0,0 +1,671 @@
...
+
+Here we will describe how `cc-runtime` handles the most important OCI commands.
+
+#### [`create`](https://github.com/clearcontainers/runtime/blob/master/create.go)
+
+When handling the OCI `create` command, `cc-runtime` goes through the following steps:
+
+1. Create the networking container namespace on the host, according to the container
+  OCI configuration file. We only support networking namespaces for now, but
+  will support more of them later.
```
grahamwhaley: s/of them/namespaces/

grahamwhaley: or even s/more of them/other namespaces/

docs/architecture/architecture.md
```diff
@@ -0,0 +1,671 @@
...
+2. Run all the [prestart OCI hooks](https://github.com/opencontainers/runtime-spec/blob/master/config.md#hooks)
+  in the host namespaces created in step 1, as described by the OCI container
+  configuration file.
+3. [Set up the container networking namespace up](#networking). This is when
```
grahamwhaley: 'Set up the container networking namespace up': Is that 'up' at the end a typo/redundant?

docs/architecture/architecture.md
```diff
@@ -0,0 +1,671 @@
...
+  all networking interfaces are created and setup inside the namespace, according
+  to the selected pod networking model (CNM for Docker or CNI for Kubernetes).
+4. Create and start the virtual machine running the container process. The VM
```
grahamwhaley: This reads like we start and run the VM and container process(workload) here. I think we actually run the workload process in 'start', so maybe better to:
s/running/that will run/

docs/architecture/architecture.md
```diff
@@ -0,0 +1,671 @@
...
+  will run inside the host namespaces created during step 1, and its `systemd`
+  instance will spawn the `cc-agent` daemon.
+5. Register the virtual machine with `cc-proxy`.
+6. The `cc-proxy` waits for the agent to signal that it is ready and then returns
+  a token. This token uniquely identifies a process within a container inside
+  the virtual machine.
+7. Spawn the `cc-shim` process providing two arguments:
+  `cc-shim --token $(token) --uri $(uri)`
+   * The proxy URL, which can be either a UNIX or a TCP socket.
```
grahamwhaley: as the argument name to the shim is '--uri', maybe we should say 'proxy URI' just to be consistent.

docs/architecture/architecture.md
```diff
@@ -0,0 +1,671 @@
...
+
+At this point, the virtual machine that will run the containers workloads
+is up and running, and the [`cc-agent`](#agent) is ready to process container
+life cycle commands. The pod inside the virtual machine is not created, and
+the containers are not running yet. This will be done through the OCI 
+[`start`](#start) command
+
+#### [`start`](https://github.com/clearcontainers/runtime/blob/master/start.go)
+
+With namespace containers `start` launches a traditional Linux container process
```
grahamwhaley: I'd put a ',' in after containers -> 'With namespace containers, `start`...'  Will also be more consistent with the sentence that follows.

docs/architecture/architecture.md
```diff
@@ -0,0 +1,671 @@
...
+In practice, this means `cc-runtime` will run through the following steps:
+
+1. `cc-runtime` connects to `cc-proxy` and sends it the `attach` command to
+  let it know which VM we want to use to create a pod and start the new container.
+2. `cc-runtime` sends an agent `STARTPOD` command via `cc-proxy`. This creates
+  the container pod.
+3. `cc-runtime` sends an agent `NEWCONTAINER` command via `cc-proxy` in order to
+  create and start a new container in a given pod. The command is sent to `cc-proxy`
+  who forwards it to the right agent instance running in the appropriate guest.
+
```
grahamwhaley: Just note - we don't mention the pause items here, but they appear in the state diagram.
It is also not clear if the container workload is actually executed by this command or not - we could add a small note to clarify that.

docs/architecture/architecture.md
```diff
@@ -0,0 +1,671 @@
...
+![Docker start](arch-images/start.png)
+
+#### [`exec`](https://github.com/clearcontainers/runtime/blob/master/exec.go)
+
+OCI `exec` allows you to run an additional command within an already running
+container. With Clear Containers, this translates into sending a `EXECCMD` command
+to the agent so that it runs a command into a running container belonging to a
```
grahamwhaley: s/it runs a command into/executes a command in/

docs/architecture/architecture.md
```diff
@@ -0,0 +1,671 @@
...
+
+The `exec` code path is partly similar to the `create` one and `cc-runtime`
+goes through the following steps:
+
+1. `cc-runtime` connects to `cc-proxy` and sends it the `attach` command to let
+  it know which VM to run the `exec` command.
+2. `cc-runtime` receives a token from `cc-proxy` based on this connection.
+3. `cc-runtime` spawns a `cc-shim` process providing two arguments:
+  `cc-shim --token $(token) --uri $(uri)`
+   * The proxy URL, which can be either a UNIX or a TCP socket.
```
grahamwhaley: potentially s/URL/URI/ again

docs/architecture/architecture.md
```diff
@@ -0,0 +1,671 @@
...
+2. `cc-runtime` sends an agent `KILLCONTAINER` command to `kill` the container
+  running on the guest. The command is sent to `cc-proxy` who forwards it to the
+  right agent instance running in the appropriate guest.
+
+After step #2, if the container process terminates in the VM, `cc-proxy` will
+forward the process exit code to the `shim`. The `shim` will then disconnect
+from `cc-proxy` and terminates as well.
+
+One corner case `cc-runtime` also needs to handle is when it receives a signal
+for a container that's not yet running. In that case, `cc-runtime` simply stops
```
grahamwhaley: This reads a little strange to me. Let's see if I can clarify (and somebody might want to check it is still accurate...)
'In that case, for the `SIGTERM` or `SIGKILL` signals, `cc-runtime` will kill the `cc-shim`. All other signals are ignored.'

docs/architecture/architecture.md
```diff
@@ -0,0 +1,671 @@
...
+asked to. In that case it will first `kill` the container and only then `delete`
+it.
+
+The `delete` code path differs significantly between having to delete one container
+inside a pod and having to delete an entire pod. In the former case, `cc-runtime`
+will only send a `SIGKILL` signal to the container process while the latter case is
+about completely deleting the whole thing: the pod, its containers, all `cc-shim`
+instances and the virtual machine itself.
+Also, the latter case is only supported by Kubernetes whereas Docker always asks
+for the former one.
```
grahamwhaley: This makes it sound slightly like Docker will never ask for the cc-shim and VM to be deleted - I don't know if we nee to clarify that a little here.

docs/architecture/architecture.md
```diff
@@ -0,0 +1,671 @@
...
+`cc-proxy` is a daemon offering access to the VM [`cc-agent`](https://github.com/clearcontainers/agent)
+to multiple `cc-shim` and `cc-runtime` clients. Only a single instance of `cc-proxy`
+per host is necessary as it can be used for several different VMs.
+Its main role is to:
+- Arbitrate access to the `cc-agent` control channel between all the `cc-runtime`
+  instances and the `cc-shim` ones.
+- Route the I/O streams and signals between the various `cc-shim` instances and
+  the `cc-agent`.
+
+`cc-proxy` API is available through a single socket for all `cc-shim` and `cc-runtime`
```
grahamwhaley: Insert a 'The' at the start of the sentence? - 'The `cc-proxy` API...'

docs/architecture/architecture.md
```diff
@@ -0,0 +1,671 @@
...
+  virtual machine. When registering a new VM, `cc-runtime` will ask `cc-proxy`
+  for a token and pass that token to the `cc-shim` instance it will eventually
+  create for handling the container process inside that VM.
+- `UnregisterVM`: Does the opposite of what `RegisterVM` does, indicating to
+  the proxy it should release resources created by `RegisterVM`.
+- `AttachVM`: It can be used to associate clients to an already known VM.
+  Optionally `AttachVM` senders can ask `cc-proxy` for a token. In the OCI `exec`
+  command case, this token will be used by the `cc-shim` instance that monitors
+  the executed process inside the container.
+- `Hyper` : This payload will forward an hyperstart command to the `cc-agent`.
```
grahamwhaley: extraneous space ' ' after `Hyper`, before the :

grahamwhaley: s/an hyperstart/a hyperstart/

docs/architecture/architecture.md
```diff
@@ -0,0 +1,671 @@
...
+- `ConnectShim`: Used by `cc-shim` instances to connect to `cc-proxy` and let
+  it know which container process it wants to monitor. `cc-shim` sends the token
+  they've been given by `cc-runtime` as part of the `ConnectShim` payload. With
+  that information, `cc-proxy` builds its I/O and signal multiplexing routes
+  between the VM process containers and the host `cc-shim` instances.
+- `DisconnectShim`: Used by `cc-shim` instances to disconnect themselves from
+  `cc-proxy` and let it know that they stop monitoring their container process.
+
+For more details about `cc-proxy`'s protocol, theory of operations or debugging
+tips, please read
```
grahamwhaley: s/read/read the/

docs/architecture/architecture.md
```diff
@@ -0,0 +1,671 @@
...
+[`cc-proxy` README](https://github.com/clearcontainers/proxy) or the
+[proxy api `godoc`](https://godoc.org/github.com/clearcontainers/proxy/api).
+
+## Shim
+
+A container process reaper, such as Docker's `containerd-shim` or crio's `conmon`,
```
grahamwhaley: s/crios/CRI-O's/

docs/architecture/architecture.md
```diff
@@ -0,0 +1,671 @@
...
+monitor a process running within a virtual machine. At most it can see the QEMU
+process, but that is not enough. With Clear Containers, `cc-shim` acts as the
+container process that the container process reaper can monitor. Therefore
+`cc-shim` needs to handle all container I/O streams (`stdout`, `stdin` and `stderr`)
+and forward all signals the container process reaper decides to send to the container
+process.
+
+`cc-shim` has an implicit knowledge about which VM agent will handle those streams
+and signals and thus acts as an encapsulation layer between the container process
+reaper and the `cc-agent`. `cc-shim`:
```
grahamwhaley: Maybe put `cc-shim`: on a new paragraph, and expand a little to something like 'cc-shim performs the following steps:' or similar.

docs/architecture/architecture.md
```diff
@@ -0,0 +1,671 @@
...
+
+- Connects to `cc-proxy` using a token obtained by calling the `cc-proxy` `ConnectShim`
+  command. The token is passed from `cc-runtime` to `cc-shim` when the former
+  spawns the latter and is used to identify the true container process that the
+  shim process will be shadowing or representing.
+- Fragments and encapsulates the standard input stream from the container process
+  reaper into `cc-proxy` stream frames:
+```
+                  1 1 1 1 1 1 1 1 1 1 2 2 2 2 2 2 2 2 2 2 3 3
```
grahamwhaley: As I think @egernst and @amshinde have noted - the '1 1 1 1 ' line seems redundant and makes no sense. Yes, we queried this on the original doc, and I have a feeling the answer was 'oops, a copy/paste issue' :-)

grahamwhaley: I should note - if anything then I think it would have been denoting byte lanes - but it is failing at that on many fronts right now. It might be nice to show it as byte lanes though (`1 1 1 1 1 1 1 1 2 2 2 2 2 2 2 2 3 3 3 3 3 3 3 3`

docs/architecture/architecture.md
```diff
@@ -0,0 +1,671 @@
...
+docker cp [OPTIONS] HOST:SRC_PATH CONTAINER:DEST_PATH
+```
+
+Clear Containers has the ability to hotplug block devices, which makes it
+possible to use block devices for containers started after the VM has been launched.
+
+Users can check to see if the container uses devicemapper block device as its
+rootfs by calling `mount(8)` within the counter.  If devicemapper block device
+is used, '/' will be mounted on `/dev/vda`.
+
```
grahamwhaley: We could note that devicemapper block device mode can be disabled in the runtime config file if necessary.

grahamwhaley:
This is looking really good - a visible improvement on the already rather good CC2.x document - so, kudos to everybody involved in the update.

docs/architecture/architecture.md
```diff
@@ -0,0 +1,671 @@
...
+    * [DAX](#dax)
+    * [Previous Releases](#previous-releases)
+    * [Resources](#resources)
+
+## Overview
+
+This is an architectural overview of Clear Containers, based on the 3.0 release.
+
+The [Clear Containers runtime (cc-runtime)](https://github.com/clearcontainers/runtime)
+is compatible with the [OCI](https://github.com/opencontainers) [runtime specification](https://github.com/opencontainers/runtime-spec)
```
jodh-intel: Hi @iphutch - does OCI need an asterisk here?

docs/architecture/architecture.md
```diff
@@ -0,0 +1,671 @@
...
+  does not waste time initializing devices of no use for containers.
+- Skipped the guest BIOS/firmware and jumped straight to the Clear Containers
+  kernel.
+
+In order to provide some advanced features which were lacking in `pc-lite` like hotplug
+support, the default was changed to `pc`.
+
+In the future, Clear Containers plan to move to a 2.9 based version of QEMU,
+available at 
+the [Clear Containers QEMU repo](https://github.com/clearcontainers/qemu/tree/qemu-lite-v2.9.0).
```
jodh-intel: s/repo/repository/.

docs/architecture/architecture.md
```diff
@@ -0,0 +1,671 @@
...
+look to transition to this machine type.
+ 
+## Agent
+
+[`cc-agent`](https://github.com/clearcontainers/agent) is a daemon running in the
+guest as a supervisor for managing containers and processes potentially running
+within those containers.
+
+The `cc-agent` execution unit is the pod. A `cc-agent` pod is a container sandbox
+defined by a set of namespaces (NS, UTS, IPC and PID). `cc-runtime` can run several
```
jodh-intel: s/NS/mount, net, cgroup/ I think @sameo?

docs/architecture/architecture.md
```diff
@@ -0,0 +1,671 @@
...
+
+In order to provide some advanced features which were lacking in `pc-lite` like hotplug
+support, the default was changed to `pc`.
+
+In the future, Clear Containers plan to move to a 2.9 based version of QEMU,
+available at 
+the [Clear Containers QEMU repo](https://github.com/clearcontainers/qemu/tree/qemu-lite-v2.9.0).
+This transition has been delayed until after the release of Clear Containers 3.0
+due to regressions, as described in [runtime issue 407](https://github.com/clearcontainers/runtime/issues/407).
+Once support for features like hotplug are available in `Q35`, the project will
```
jodh-intel: s/Q35/q35/ for consistency with all other mentions of machine types.

docs/architecture/architecture.md
egernst: done

docs/architecture/architecture.md
```diff
@@ -0,0 +1,671 @@
...
+[Docker\* Engine](https://www.docker.com/products/docker-engine) pluggable runtime
+architecture. It also supports the [Kubernetes\* Container Runtime Interface (CRI)](https://github.com/kubernetes/kubernetes/tree/master/pkg/kubelet/apis/cri/v1alpha1/runtime)
+through the [CRI-O\*](https://github.com/kubernetes-incubator/cri-o) implementation.
+In other words, you can transparently select between the
+[default Docker and CRI-O runtime (runc)](https://github.com/opencontainers/runc)
+and `cc-runtime`.
+
+![Docker and Clear Containers](arch-images/docker-cc.png)
+
+`cc-runtime` creates a QEMU\*/KVM virtual machine for each container the Docker
```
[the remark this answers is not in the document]
egernst: Sure, agreed.

docs/architecture/architecture.md
```diff
@@ -0,0 +1,671 @@
...
+library.
+
+### Configuration
+
+The runtime uses a TOML format configuration file called `configuration.toml`. By
+default this file is installed in the `/etc/clear-containers` directory.
+Most users will not need to modify the configuration file.
+
+The file is well commented and provides a few "knobs" that can modify the
+behaviour of the runtime.
```
jodh-intel: I wrote this but on re-reading, I think it would be better to say, "that can be used to modify the behaviour of the runtime."

docs/architecture/architecture.md
```diff
@@ -0,0 +1,671 @@
...
+
+### CNM
+
+![High-level CNM Diagram](arch-images/CNM_overall_diagram.png)
+
+__CNM lifecycle__
+
+1.  RequestPool
+
+2.  CreateNetwork
```
jodh-intel: These should all be rendered in a fixed-format using backticks.

docs/architecture/architecture.md
```diff
@@ -0,0 +1,671 @@
...
+
+3.  RequestAddress
+
+4.  CreateEndPoint
+
+5.  CreateContainer
+
+6.  Create config.json
```
jodh-intel: `config.json` (same for below).

docs/architecture/architecture.md
egernst: done

docs/architecture/architecture.md
```diff
@@ -0,0 +1,671 @@
...
+
+Clear Containers supports both
+[CNM](https://github.com/docker/libnetwork/blob/master/docs/design.md#the-container-network-model)
+and [CNI](https://github.com/containernetworking/cni) for networking management.
+
+### CNM
+
+![High-level CNM Diagram](arch-images/CNM_overall_diagram.png)
+
+__CNM lifecycle__
```
jodh-intel: I think this needs an introductory sentence or two - currently it's just a list of words.

docs/architecture/architecture.md
```diff
@@ -0,0 +1,671 @@
...
+![CNI Diagram](arch-images/CNI_diagram.png)
+
+__Runtime network setup with CNI__
+
+1. Create the network namespace
+
+2. Get CNI plugin information
+
+3. Start the plugin (providing previously created netns) to add a network
+  described into ```/etc/cni/net.d/ directory```. At that time, the CNI plugin will
```
jodh-intel: Triple-backticks but only single ones required.

docs/architecture/architecture.md
egernst: nice, done

docs/architecture/architecture.md
egernst: agreed. done

docs/architecture/architecture.md
```diff
@@ -0,0 +1,671 @@
...
+  create the cni0 network interface and a veth pair between the host and the created
+  netns. It links cni0 to the veth pair before to exit.
```
jodh-intel: I don't know what @iphutch feels about this, but I'd rather we were consistent with periods: either all list entries end with one, or none do.

docs/architecture/architecture.md
```diff
@@ -0,0 +1,671 @@
...
+
+__Runtime network setup with CNI__
+
+1. Create the network namespace
+
+2. Get CNI plugin information
+
+3. Start the plugin (providing previously created netns) to add a network
+  described into ```/etc/cni/net.d/ directory```. At that time, the CNI plugin will
+  create the cni0 network interface and a veth pair between the host and the created
```
jodh-intel: s/cni0/`cni0`/.

docs/architecture/architecture.md
```diff
@@ -0,0 +1,671 @@
...
+host to container accesses the mounted file system on the host side. This is
+not expected to work and may lead to inconsistencies as the block device will
+be simultaneously written to, from two different mounts. The copy operation from
+container to host will work, provided the user calls `sync(1)` from within the
+container prior to the copy to make sure any outstanding cached data is written
+to the block device.
+
+```
+docker cp [OPTIONS] CONTAINER:SRC_PATH HOST:DEST_PATH
+docker cp [OPTIONS] HOST:SRC_PATH CONTAINER:DEST_PATH
```
jodh-intel: This is a generic "usage statement" and is hard to read given it is in upper-caps (well, it is for me ;). I think a concrete example would be helpful here.

docs/architecture/architecture.md
egernst: done

docs/architecture/architecture.md
```diff
@@ -0,0 +1,671 @@
...
+```
+
+Clear Containers has the ability to hotplug block devices, which makes it
```
jodh-intel: Mention that devices can be added **and** removed? Also, shouldn't this be "Intel® Clear Containers" (here and elsewhere)?

docs/architecture/architecture.md
egernst: done

docs/architecture/architecture.md
```diff
@@ -0,0 +1,671 @@
...
+possible to use block devices for containers started after the VM has been launched.
+
+Users can check to see if the container uses devicemapper block device as its
```
jodh-intel: s/uses devicemapper/uses the devicemapper/

docs/architecture/architecture.md
```diff
@@ -0,0 +1,671 @@
...
+rootfs by calling `mount(8)` within the counter.  If devicemapper block device
```
jodh-intel: s/If devicemapper/If the devicemapper/

docs/architecture/architecture.md
```diff
@@ -0,0 +1,671 @@
...
+is used, '/' will be mounted on `/dev/vda`.
```
jodh-intel: Replace '/' with "the root directory, `/`, "

docs/architecture/architecture.md
```diff
@@ -0,0 +1,671 @@
...
+
+# Appendices
+
+## DAX
+
+Clear Containers utilises the Linux kernel DAX [(Direct Access filesystem)](https://git.kernel.org/cgit/linux/kernel/git/torvalds/linux.git/tree/Documentation/filesystems/dax.txt)
+feature to efficiently map some host side files into the guest VM space.
```
jodh-intel: @iphutch - should "host side" be hyphenated here?

docs/architecture/architecture.md
egernst: ack

docs/architecture/architecture.md
egernst: ack

docs/architecture/architecture.md
```diff
@@ -0,0 +1,671 @@
...
+## Agent
+
+[`cc-agent`](https://github.com/clearcontainers/agent) is a daemon running in the
+guest as a supervisor for managing containers and processes potentially running
+within those containers.
+
+The `cc-agent` execution unit is the pod. A `cc-agent` pod is a container sandbox
+defined by a set of namespaces (NS, UTS, IPC and PID). `cc-runtime` can run several
+containers per pod to support container engines that require multiple containers
+running inside a single VM. In the case of docker, `cc-runtime` creates a single
```
jodh-intel: We haven't defined what "VM" means. I suggest we add a Terminology section for this and other such terms.

docs/architecture/architecture.md
egernst: this reads much better to me, good suggestion.

docs/architecture/architecture.md
```diff
@@ -0,0 +1,671 @@
...
+
+![Detailed CNM Diagram](arch-images/CNM_detailed_diagram.png)
+
+__Runtime network setup with CNM__
+
+1. Read config.json
+
+2. Create the network namespace
+
+3. Call the prestart hook (from inside the netns)
```
jodh-intel: The term "netns" needs to be defined in a Terminology section.

docs/architecture/architecture.md
jodh-intel: We are not being consistent here: sometimes stdin/out/err are rendered with backticks, sometimes not (I'd opt for always using backticks for these three terms).

docs/architecture/architecture.md
```diff
@@ -0,0 +1,671 @@
...
+4. Create bridge, TAP, and link all together with network interface previously
+  created
+
+5. Start VM inside the netns and start the container
+
+## Storage
+Container workloads are shared with the virtualized environment through [9pfs](https://www.kernel.org/doc/Documentation/filesystems/9p.txt).
+The devicemapper storage driver is a special case. The driver uses dedicated block
+devices rather than formatted filesystems, and operates at the block level rather than the file
+level. This knowledge has been used to directly use the underlying block device
```
jodh-intel: s/has been used/is used/

docs/architecture/architecture.md
```diff
@@ -0,0 +1,671 @@
...
+instead of the overlay file system for the container root file system. The block
+device maps to the top read-write layer for the overlay. This approach gives much
+better I/O performance compared to using 9pfs to share the container file system.
+
+The approach above does introduce a limitation in terms of dynamic file copy
+in/out of the container via `docker cp` operations. The copy operation from
```
jodh-intel: I like the word "via" but for non-native speakers maybe "using the" (or similar) would be better.

docs/architecture/architecture.md
```diff
@@ -0,0 +1,671 @@
...
+host to container accesses the mounted file system on the host side. This is
+not expected to work and may lead to inconsistencies as the block device will
+be simultaneously written to, from two different mounts. The copy operation from
```
jodh-intel: I don't think this comma is required.

docs/architecture/architecture.md
```diff
@@ -0,0 +1,671 @@
...
+1. Create the network namespace
+
+2. Get CNI plugin information
+
+3. Start the plugin (providing previously created netns) to add a network
+  described into ```/etc/cni/net.d/ directory```. At that time, the CNI plugin will
+  create the cni0 network interface and a veth pair between the host and the created
+  netns. It links cni0 to the veth pair before to exit.
+
+4. Create bridge, TAP, and link all together with network interface previously
```
jodh-intel: Maybe "network bridge" (here and elsewhere) and a definition of TAP in the Terminology section?

docs/architecture/architecture.md
egernst: @grahamwhaley @sameo @jodh-intel - I agree, I don't like how it is worded as is.  I was moving towards:
```
328 The `delete` code path differs significantly between having to delete one container
329 inside a pod (as is typical in Docker) and having to delete an entire pod (which
330 is unique to Kubernetes). In the former case, `cc-runtime` will only send a   
331 `SIGKILL` signal to the container process.  In the latter case, the whole thing 
332 is deleted: the pod, its containers, all `cc-shim` instances and the virtual  
333 machine itself.    
```
--but this still doesn't really address your point.  Suggestions?

docs/architecture/architecture.md
```diff
@@ -0,0 +1,671 @@
...
+  created
+
+5. Start VM inside the netns and start the container
+
+## Storage
+Container workloads are shared with the virtualized environment through [9pfs](https://www.kernel.org/doc/Documentation/filesystems/9p.txt).
```
jodh-intel: s/through 9pfs/using the 9pfs filesystem/

docs/architecture/architecture.md
egernst: ack

docs/architecture/architecture.md
```diff
@@ -0,0 +1,671 @@
...
+is used, '/' will be mounted on `/dev/vda`.
+
+# Appendices
+
+## DAX
+
+Clear Containers utilises the Linux kernel DAX [(Direct Access filesystem)](https://git.kernel.org/cgit/linux/kernel/git/torvalds/linux.git/tree/Documentation/filesystems/dax.txt)
+feature to efficiently map some host side files into the guest VM space.
+In particular, Clear Containers uses the `QEMU` nvdimm feature to provide a
+memory mapped virtual device that can be used to DAX map the virtual machine's
```
jodh-intel: hyphenate "memory-mapped"?

docs/architecture/architecture.md
egernst: ack

docs/architecture/architecture.md
```diff
@@ -0,0 +1,671 @@
...
+root filesystem into the guest space.
```
jodh-intel: s/guest space/guest memory address space/

docs/architecture/architecture.md
egernst: ack

docs/architecture/architecture.md
```diff
@@ -0,0 +1,671 @@
...
+file and device mapping mechanisms:
+
+- Mapping as a direct access devices allows the guest to directly access
+  the host memory pages (such as via eXicute In Place (XIP)), bypassing the guest
+  page cache. This provides both time and space optimisations.
+- Mapping as a direct access device inside the VM allows pages from the
+  host to be demand loaded using page faults, rather than having to make requests
+  via a virtualised device (causing expensive VM exits/hypercalls), thus providing
+a speed optimisation.
+- Utilising MAP_SHARED shared memory on the host allows the host to efficiently
```
jodh-intel: s/MAP_SHARED/`MAP_SHARED`/

docs/architecture/architecture.md
egernst: ack

docs/architecture/architecture.md
```diff
@@ -0,0 +1,671 @@
...
+rootfs by calling `mount(8)` within the counter.  If devicemapper block device
+is used, '/' will be mounted on `/dev/vda`.
+
+# Appendices
+
+## DAX
+
+Clear Containers utilises the Linux kernel DAX [(Direct Access filesystem)](https://git.kernel.org/cgit/linux/kernel/git/torvalds/linux.git/tree/Documentation/filesystems/dax.txt)
+feature to efficiently map some host side files into the guest VM space.
+In particular, Clear Containers uses the `QEMU` nvdimm feature to provide a
```
jodh-intel: Remove backticks around QEMU.

docs/architecture/architecture.md
egernst: Now that it is adjusted, it is showing bit number, the 1's aren't redundant, they are the decimal MSB for each bit.  Turn your head clockwise when you look at it :)

docs/architecture/architecture.md
egernst: sure

docs/architecture/architecture.md
egernst: done.

grahamwhaley:
I think we ( @iphutch ) may have mentioned the high-level-overview.png is not referenced anywhere - to note, it was referenced from the proxy README.md in the cc2.x repo, and thus I believe has been un-necessarily carried over (and there is a copy over in the clearcontainers/proxy/docs dir I think). Yes, it can be dropped from here.

docs/architecture/architecture.md
```diff
@@ -0,0 +1,671 @@
...
+#### [`delete`](https://github.com/clearcontainers/runtime/blob/master/delete.go)
+
+`delete` is about deleting all resources held by a stopped/killed container.
+Running containers can not be deleted unless the OCI runtime is explictly being
+asked to. In that case it will first `kill` the container and only then `delete`
+it.
+
+The `delete` code path differs significantly between having to delete one container
+inside a pod (as is typical in Docker) and having to delete an entire pod (which
+is unique to Kubernetes). In the former case, `cc-runtime` will only send a
```
grahamwhaley: rather than 'is unique to' I might say 'such as from'.

docs/architecture/architecture.md
```diff
@@ -0,0 +1,671 @@
...
+`SIGKILL` signal to the container process.  In the latter case, the whole thing
```
grahamwhaley: s/whole thing/whole pod/  or something similar - 'all components' or something

docs/architecture/architecture.md
iphutch: Yes, we will need an asterisk here. Good catch.

docs/architecture/architecture.md
iphutch: First instance on any page, it should be Intel® Clear Containers, however, both in the current and in the new website the full name/trademarks are in the heading or at least present somewhere on every page. This means asterisks and the full Intel® Clear Containers name isn't necessary in the body of any of our docs.

docs/architecture/architecture.md
iphutch: I generally try to avoid hyphenation but I'm on board with memory-mapped and host-side.

docs/architecture/architecture.md
iphutch: Agreed. All or none.

docs/architecture/architecture.md
iphutch: +1

docs/architecture/architecture.md
iphutch: +1

eadamsintel:
I recommend adding a Kubernetes overall architecture guide including the shim/proxy/agent and how it works with crio.

PR #544, opened 2017-09-12
build: Ensure version details correct for system build

### LINKED ISSUE #373: On building and installing with CC_SYSTEM_BUILD=yes, cc-runtime still looks for images under /usr/local/share instead of /usr/share
build output shows assets path as /usr/share/defaults, but the binary looks for /usr/local/share instead.

jodh-intel:
Requesting a system build does not correctly invalidate any previous
build.

Force the correct behaviour by ensuring a non-installation build
depends on "clean". This guarantees the generated code will be
recreated and ensures the correct version details in the resulting
binary.

Fixes #373.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -42,6 +42,8 @@ TARGET = cc-runtime
 DESTDIR :=
 CCDIR := clear-containers
 
+installing = $(findstring install,$(MAKECMDGOALS))
+
 ifeq ($(cc_system_build),yes)
     # Configure the build for a standard Clear Containers system that is
     # using OBS-generated packages.
@@ -51,6 +53,12 @@ ifeq ($(cc_system_build),yes)
     QEMUBINDIR    := $(BINDIR)
     SYSCONFDIR    := /etc
     LOCALSTATEDIR := /var
+
+    ifeq (,$(installing))
+        # Force a rebuild to ensure version details are correct
+        # (but only for a non-install build phase).
+        EXTRA_DEPS = clean
+    endif
 else
     PREFIX        := /usr/local
     BINDIR        := $(PREFIX)/bin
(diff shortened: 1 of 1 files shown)


jodh-intel:
Hi @amshinde - please can you test the behaviour on this PR.

Makefile
```diff
@@ -42,6 +42,8 @@ TARGET = cc-runtime
 DESTDIR :=
 CCDIR := clear-containers
 
+installing = $(findstring install,$(MAKECMDGOALS))
```
grahamwhaley: voodoo!

Makefile
```diff
@@ -51,6 +53,12 @@ ifeq ($(cc_system_build),yes)
     QEMUBINDIR    := $(BINDIR)
     SYSCONFDIR    := /etc
     LOCALSTATEDIR := /var
+
+    ifeq (,$(installing))
```
grahamwhaley: Forcing a clean for a non-install system build feels a little 'heavy handed', but I suspect any other better solution (such as removing, and thus forcing re-generation of, the generated file that holds the VERSION info) would require us holding some sort of 'status' file tracking that we've changed what sort of build we are doing.
Given the system-build steps are pretty well defined (build, and then install), I think this is fine as is. If we get a few extra clean/build cycles for people doing dev work, so be it.

Makefile
```diff
@@ -51,6 +53,12 @@ ifeq ($(cc_system_build),yes)
...
+        # Force a rebuild to ensure version details are correct
+	# (but only for a non-install build phase).
```
grahamwhaley: mix of tabs and spaces on these two comment lines

Makefile
jodh-intel: I agree. Yes, I guess we could just blat the generated file but I'm open to other ideas if anyone has any?

Makefile
jodh-intel: Fixed.

PR #546, opened 2017-09-12
Add upgrading doc

jodh-intel:


jodh-intel:
Hi @iphutch, @grahamwhaley, @gorozco1 - please could you review?

docs/upgrading.md
```diff
@@ -0,0 +1,111 @@
...
+  of Clear Containers which uses a runtime called `cc-oci-runtime` and
+  whose code is available at: https://github.com/01org/cc-oci-runtime
+
+- Clear Container 3.x (or just 3.x) refers to the current implementation
+  which uses a runtime called `cc-runtime` and whose code is available
+  at: https://github.com/clearcontainers/runtime
+
+## Maintenance warning
+
+The 2.x codebase is no longer being developed. New releases will _only_ be
```
grahamwhaley: nitpick: MD query: here you use _underbars_ which come out as italic, but below you use *stars* that render the same - are they notionally different?

docs/upgrading.md
```diff
@@ -0,0 +1,111 @@
...
+Assuming the system was installed by following the [CC 2.x Ubuntu installation guide](https://github.com/01org/cc-oci-runtime/blob/master/documentation/Installing-Clear-Containers-on-Ubuntu.md):
+
+1. Perform the following cleanup steps:
+   ```bash
+   $ sudo rm -f /etc/systemd/system/docker.service.d/clr-containers.conf
+   $ sudo apt-get purge -y docker-engine
+   $ sudo rm -f /etc/apt/sources.list.d/cc-oci-runtime.list
+   ```
+
+1. Follow the [CC 3.x Ubuntu installation guide](docs/ubuntu-installation-guide.md).
```
grahamwhaley: I think you might not need the pre-pended `docs` dir on this .md file url - it does not hot-link (is broken) for me, even though that file does exist in your branch. I see a 'docs/docs/' in the URL.
Same for other links (Fedora). Hopefully easy for you to check locally?

docs/upgrading.md
```diff
@@ -0,0 +1,111 @@
...
+   ```
+
+1. Follow the [CC 3.x Fedora installation guide](docs/fedora-installation-guide.md).
+
+## Unsupported operations and configurations
+
+### CentOS
+
+Since the CentOS installation instructions require the user to manually
+build and install various packages, upgrading is not supported on this
```
grahamwhaley: Rather than say 'not supported', maybe we can say 'must be performed by hand' (nee: 'left as an exercise for the user' ;-)). That is, it is not that you cannot upgrade, it is just there is no nicely defined procedure to do so.

docs/upgrading.md
```diff
@@ -0,0 +1,111 @@
...
+elements. Namely, both use a shim process called `cc-shim` and a proxy
+process called `cc-proxy`. However, the 2.x versions of these
+elements *are incompatible* with their 3.x counterparts.
+
+## Supported operations and configurations
+
+### Systems using pre-built packages
+
+For operating systems that pre-built packages are provided for (such as
+Fedora and Ubuntu), upgrading is possible with some caveats:
```
grahamwhaley: Do we need `*`'s on first mention of the tradenames such as Fedora, Ubuntu, CentOS - I suspect @iphutch will pick that up otherwise ;-)

docs/upgrading.md
```diff
@@ -0,0 +1,111 @@
...
+
+### Downgrading
+
+Systems using packaged versions of 3.x cannot be downgraded to 2.x
+simply by following the 2.x installation instructions.
+
+However, it should be possible to switch back to 2.x by first fully
+removing 3.x and then following the 2.x installation instructions.
+
+Using 2.x is now discouraged given that it is no longer being maintained
```
grahamwhaley: Nit: it is being maintained (critical fixes), but not developed (no new features) - maybe just semantics - let's check for the correct wording etc.

docs/developers-clear-containers-install.md
```diff
@@ -36,8 +36,9 @@ Since the installation guide will have installed packaged versions of
 all required components, it is only necessary to install the source for
 the component(s) you wish to develop with.
 
-**IMPORTANT:** Do not combine [Clear Containers 2.1](https://github.com/01org/cc-oci-runtime) and [Clear Containers 3.0](https://github.com/clearcontainers).
-Both projects ship ``cc-proxy`` and they are not compatible with each other.
+**IMPORTANT:** Do not combine [Clear Containers 2.x](https://github.com/01org/cc-oci-runtime) and [Clear Containers 3.x](https://github.com/clearcontainers) on the same system.
```
grahamwhaley: Thanks for doing those @jodh-intel  I've been trying to push the 'n.x' notation to reduce the number version rot...
There are places where hard versions probably do make sense (some install or arch docs for instance might be specific) - but they are probably less common than we currently carry in the docs.

docs/upgrading.md
jodh-intel: Fixing...

docs/upgrading.md
jodh-intel: Fixing...

docs/upgrading.md
jodh-intel: Not that easy to check locally, no, but fixing...

docs/upgrading.md
jodh-intel: Glad you raised this because there are 2 issues here: upgrading the system, but also whether the resulting system should be considered "supported" (in that we will consider bugs raised on such a system). I agree that we don't want to specify how to upgrade a CentOS system, but the question comes down to, "what if someone does it?" If they do, I think we say "not supported" but what do you think @sameo?

docs/upgrading.md
jodh-intel: Fixing...

jodh-intel:
@grahamwhaley - branch updated.

egernst:
@mcastelino -- FYI.

docs/upgrading.md
```diff
@@ -0,0 +1,111 @@
...
+place.
+
+#### Upgrading an Ubuntu system
+
+Assuming the system was installed by following the [CC 2.x Ubuntu installation guide](https://github.com/01org/cc-oci-runtime/blob/master/documentation/Installing-Clear-Containers-on-Ubuntu.md):
+
+1. Perform the following cleanup steps:
+   ```bash
+   $ sudo rm -f /etc/systemd/system/docker.service.d/clr-containers.conf
+   $ sudo apt-get purge -y docker-engine
```
mcastelino: @jodh-intel any reason you do not also reccomend apt-get remove and dnf remove of the packages that got isntalled

docs/developers-clear-containers-install.md
```diff
@@ -36,8 +36,9 @@ Since the installation guide will have installed packaged versions of
...
+Both projects ship ``cc-proxy`` and ``cc-shim`` which are not compatible with each other.
```
iphutch: Is there a reason we go back and fourth between double and singe back-ticks for literal text?
I am seeing that the majority of new docs coming through are using single `, this doc also uses single earlier in the text (line 25).

docs/upgrading.md
```diff
@@ -0,0 +1,111 @@
...
+
+This document discusses the available options for upgrading a system running Intel® Clear Containers.
+
+## Terminology
+
+There are currently two implementations of Clear Containers:
+
+- Clear Containers 2.x (or just 2.x) refers to the previous implementation
+  of Clear Containers which uses a runtime called `cc-oci-runtime` and
+  whose code is available at: https://github.com/01org/cc-oci-runtime
```
iphutch: Let's re-work this to avoid the use of "whose":
called `cc-oci-runtime`. 2.x code is available at: https://github.com/01org/cc-oci-runtime

docs/upgrading.md
```diff
@@ -0,0 +1,111 @@
...
+elements. Namely, both use a shim process called `cc-shim` and a proxy
+process called `cc-proxy`. However, the 2.x versions of these
+elements *are incompatible* with their 3.x counterparts.
+
+## Supported operations and configurations
+
+### Systems using pre-built packages
+
+For operating systems that pre-built packages are provided for (such as
+Fedora* and Ubuntu*), upgrading is possible with some caveats:
```
iphutch: replace "caveats" with "conditions" or "limitations"
replace/that pre-built packages are provided for/with pre-build packages

docs/upgrading.md
```diff
@@ -0,0 +1,111 @@
...
+
+## Introduction
+
+The two Clear Containers implementations share certain architectural
+elements. Namely, both use a shim process called `cc-shim` and a proxy
+process called `cc-proxy`. However, the 2.x versions of these
+elements *are incompatible* with their 3.x counterparts.
+
+## Supported operations and configurations
+
```
iphutch: Include a description under a section. We want to avoid having two headings one after another.

docs/upgrading.md
```diff
@@ -0,0 +1,111 @@
...
+1. Perform the following cleanup steps:
+   ```bash
+   $ sudo rm -f /etc/systemd/system/docker.service.d/clr-containers.conf
+   $ sudo rm -f /etc/yum.repos.d/home:clearlinux:preview:clear-containers-2.1.repo
+   ```
+
+1. Follow the [CC 3.x Fedora installation guide](fedora-installation-guide.md).
+
+## Unsupported operations and configurations
+
```
iphutch: Ditto

docs/upgrading.md
```diff
@@ -0,0 +1,111 @@
...
+### CentOS
```
iphutch: Need * here for the first instance of CentOS

docs/upgrading.md
```diff
@@ -0,0 +1,111 @@
...
+
+### Configuration files
+
+The 2.x runtime supported three different configuration files
+(`vm.json`, `hypervisor.args` and `cc-oci-runtime.sh.cfg` for the
+`cc-oci-runtime.sh` helper script).
+
+The 3.x runtime uses a single configuration file (`configuration.toml`).
+
+Note that none of the 2.x configuration files are migrated or read by
```
iphutch: Note:

None of the..

docs/upgrading.md
```diff
@@ -0,0 +1,111 @@
...
+the 3.x runtime. Any changes made to the 2.x configuration files will
+**not** be carried over to the 3.x runtime.
+
+### Downgrading
+
+Systems using packaged versions of 3.x cannot be downgraded to 2.x
+simply by following the 2.x installation instructions.
+
+However, it should be possible to switch back to 2.x by first fully
+removing 3.x and then following the 2.x installation instructions.
```
iphutch: lets have this "installation instructions" a link

docs/upgrading.md
jodh-intel: We don't need to do that as the packaging changes will auto-remove the old CC 2.x packages when the user installs the new ones with `apt-get`/`dnf`. However, I found in testing that docker gets upset about being upgraded hence this entry.

docs/developers-clear-containers-install.md
jodh-intel: This edit was just an update to an existing doc (containing the double-ticks). However, I agree so will push a new commit with single ticks...

jodh-intel: Done.

jodh-intel:
Hi @iphutch - thanks for reviewing. Branch updated based on your feedback.

docs/upgrading.md
jodh-intel: I think this is the only remaining issue blocking this landing. @sameo - do you have a view on this?

sameo: @jodh-intel I'd rather avoid saying "not supported" because we're not at a point where we can ignore users because they've done a manual upgrade. Realistically, if someone reports a bug on an upgraded CentOS we'll have to look at it I'm afraid.

jodh-intel: Hi @sameo, @grahamwhaley - I've updated the doc and moved the CentOS section below "Supported" -> "Not Recommended" which will hopefully make it clear ;)

Can you re-review?

grahamwhaley:
but you have the travis revision range issue rearing its head :-(

PR #578, opened 2017-09-19
.ci: Remove Travis and add unit testing to all CI

### LINKED ISSUE #577: Remove travis and run unit testing and checkcommits on all CI environments
By removing travis, we need to be able to run unit testing and checkcommits on the remaining CI system, jenkins in our case.

sboeuf:
Unit testing and checkcommits were only run against Travis. Because
we want to rely on only one CI, Jenkins in our case, the solution is
to run both unit testing and checkcommits through Jenkins too.

Fixes #577


grahamwhaley:
/me interested to see how this pans out on jenkins :-)
lgtm

grahamwhaley:
Hah, Travis failure due to lack of VMX :-) I guess we might need to guard against TRAVIS running all the checks etc.

sboeuf:
@grahamwhaley yes I know, I am on it, there are more changes to apply.

sboeuf:
@jodh-intel @grahamwhaley @sameo Please merge this.

.ci/run.sh
```diff
@@ -18,6 +18,9 @@
 
 set -e
 
+# Run unit testing
+sudo -E PATH=$PATH bash -c "make check"
```
jodh-intel: This should be just `make check` since `.ci/go-test.sh` elevates privs to `root` using `sudo` itself. Unless you drop the `sudo` here, the unit tests only get run once - as `root`, but we need them to run twice (as non-`root` and as `root`) to ensure all the unit test coverage scenarios are exercised.

grahamwhaley:
Hi @sboeuf Adding a note here, as I'm not sure there is a better place:
I've disabled (commented out) the goveralls coverage lines from the Jenkins 16.04 runtime start script, as we were seeing failures with:
```
Ginkgo ran 1 suite in 12m22.601736901s
Test Suite Passed
Error parsing coverage: open profile.cov: no such file or directory
Build step 'Execute shell' marked build as failure
```

I believe we enabled that as we are removing the Travis invocation with this PR.

.ci/run.sh
sboeuf: I can do that, but the thing is that we .ci/run.sh script is called with sudo, meaning that calling make check without sudo will end up doing the same thing. I mean that the user is gonna be `root` also in that case, right ?
What do you think we should do ?

jodh-intel: I'm not really following I'm afraid. But we must run `.ci/go-test.sh` as a non-`root` user to allow it to elevate its privileges using `sudo`.

sboeuf: I get that, but I am saying that `.ci/run.sh` is run as root. Don't you think this will prevent us from running `go-test.sh` as non root ?

jodh-intel: Well, we I'm saying we need to make `.ci/go-test.sh` run as non-root, as it does under Travis to ensure we run the tests correctly ;)

grahamwhaley: So, we currently run `sudo -E PATH=$PATH bash -c ".ci/run.sh"` in the Jenkins script - I guess the questions are:
1) Do we need to run that as sudo?
2) Should we run it maybe as both sudo and no sudo?

The default user is `USER=jenkins` according to the logs...

jodh-intel: If `.ci/run.sh` is already run as `root`, why have you changed that script to add the `sudo` call?

grahamwhaley: (crossed comments in the night....) so, if we just remove the sudo from the current run.sh line in the Jenkins script - do we see any potential problems anybody? /cc @sboeuf @chavafg

sboeuf: I don't think we need to run this as sudo since jenkins user has sudo capabilities, I will give this a try. I don't think we should run this twice from a jenkins perspective, otherwise our CI will take forever, only for unit test purpose.

jodh-intel: We're only running `go test` twice (it's fast and already handled by `.ci/go-test.sh`).

grahamwhaley:
merging, ignoring the travis and coveralls CI's as we know they are in flux/WIP.

Issue #568, opened 2017-09-18
Advanced Memory Management support


mcastelino:
Add support in the clear container runtime to support advanced memory management features like
- Memory pre-allocation support (for latency sensitive applications)
- Memory swap enabling (for density sensitive deployments), where performance requirements are relaxed.

mcastelino:
Implemented in https://github.com/clearcontainers/runtime/pull/550

mcastelino:
can be closed once https://github.com/clearcontainers/runtime/pull/550 is merged

PR #584, opened 2017-09-19
Configuration: Add support for fine grained control of memory management and debug

mcastelino:
# Add support for fine grained control of memory management.

    ```enable_prealloc=true```  Will result in the upfront allocation  and locking down of all VM memory, resulting in lower VM density.

    ```enable_swap=true```   Will enable swap of VM memory, resulting in higher VM density with potentially higher latency

Enabling the two together results in undefined behavior.

    ```enable_debug=true```   Will enable hypervisor and container kernel logs

Fixes https://github.com/clearcontainers/runtime/issues/568

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -111,6 +111,7 @@ DEFMEMSZ := 2048
 
 DEFDISABLEBLOCK := false
 DEFENABLEMEMPREALLOC := false
+DEFENABLESWAP := false
 
 SED = sed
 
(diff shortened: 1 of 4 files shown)


mcastelino:
@sameo hopefully this fixes checkcommit

jodh-intel:
> Enabling the two together results in undefined behavior.

Somebody is bound to do this, so I wonder if we should (at some point) add these new options to the output of `cc-env`. Meantime, we can ask users reporting issues to attach the output of `cc-env` along with their config file to diagnose this scenario.

lgtm.

grahamwhaley:
We can set up a 'template' for Issue reporting on github that will indicate what info to attach and how to get it... that will help us get the `cc-check` info and `docker info` attached to the Issues I believe

jodh-intel:
@grahamwhaley - #589 raised to capture this.

sameo:
I just merged PR #590, which is a rebased version of this to trigger a new jenkins build.

PR #602, opened 2017-09-20
Configuration: Provide an option to disable nesting customizations

mcastelino:
Provide an option to disable nesting customization. This is useful
when running on top of KVM, where disabling the customizations
performed when the runtime detects it is running on top of a VM
results in higher performance.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -113,6 +113,7 @@ DEFDISABLEBLOCK := false
 DEFENABLEMEMPREALLOC := false
 DEFENABLESWAP := false
 DEFENABLEDEBUG := false
+DEFDISABLENESTINGCHECKS := false
 
 SED = sed
 
@@ -187,6 +188,7 @@ USER_VARS += DEFDISABLEBLOCK
 USER_VARS += DEFENABLEMEMPREALLOC
 USER_VARS += DEFENABLESWAP
 USER_VARS += DEFENABLEDEBUG
+USER_VARS += DEFDISABLENESTINGCHECKS
 
 
 V              = @
@@ -239,6 +241,7 @@ const defaultDisableBlockDeviceUse bool = $(DEFDISABLEBLOCK)
 const defaultEnableMemPrealloc bool = $(DEFENABLEMEMPREALLOC)
 const defaultEnableSwap bool = $(DEFENABLESWAP)
 const defaultEnableDebug bool = $(DEFENABLEDEBUG)
+const defaultDisableNestingChecks bool = $(DEFDISABLENESTINGCHECKS)
 
 // Default config file used by stateless systems.
 var defaultRuntimeConfiguration = "$(DESTCONFIG)"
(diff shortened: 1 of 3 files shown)


jodh-intel:
lgtm

grahamwhaley:
lgtm
but, CI is failing as it says it cannot see a Fixes, but I see one in the commit, and the SHAs match?
```
23:22:09 Found 1 commit between commit HEAD and branch master
23:22:09 Checking commit 6b669e7d3ab492a563c23cfe50684f999a26df5c
23:22:09 No "Fixes" found
23:22:09 Build step 'Execute shell' marked build as failure
```
/cc @jodh-intel 
I'm going to tray the magic new Jenkins 'rebuild' button if I can to see if it was sporadic (although it feels like it would not be.....)

grahamwhaley:
Nudged a Jenkins rebuild for 17.04 - let's see how that goes.

grahamwhaley:
OK, the rebuild also failed. Odd, it looks like it picked up different commit SHAs, even thought the initial git checkout looks the same I think. @jodh-intel @sboeuf - maybe this has to do with the recent change to our Jenkins config that enabled building from non-master and checking against master merges?
```
10:31:24 Running checkcommits version 0.0.1 (commit f9afea96563c9296c54470eec86376c596f19f6f)
10:31:24 Defaulting commit to HEAD
10:31:24 Defaulting branch to master
10:31:24 Found 1 commit between commit HEAD and branch master
10:31:24 Checking commit 3213319062a3ea5328a97feebfe45c89d0cd38ec
10:31:24 No "Fixes" found
```
(note, that SHA 321331 is not the commited patch, nor does it match the SHA of the previous checkcommit fail, which was the commited patch ?)

jodh-intel:
@grahamwhaley - sounds like a distinct possibility. @sboeuf - do you know if the jenkins/github integration sets any useful environment variables that would allow `checkcommits` to determine the correct commit details? Travis and Semaphore both do this so I'm assuming there will be some sort of equivalent?

grahamwhaley:
Looks like we are using this github trigger plugin:
https://wiki.jenkins.io/display/JENKINS/GitHub+pull+request+builder+plugin
There are some details there of the env vars it sets in the 'Environment Variables' section.
I presume they names are self explanatory, as there is not a lot of detail on what each var actually is. The code that sets them up appears to be at :https://github.com/jenkinsci/ghprb-plugin/blob/6c4b32b6c1512314e20f7f9ac28ca99e619f4c5d/src/main/java/org/jenkinsci/plugins/ghprb/GhprbTrigger.java#L331

jodh-intel:
Well, I'd rather see actual examples of the variables. In fact, if we can find a way to separate the output of our jenkins jobs into separate output files, I think it would be worth running [procenv](https://github.com/jamesodhunt/procenv) (it's in the Ubuntu + Fedora repos) at the start of each build as all env details (including variables of course ;-) will then be available to help with debugging.

grahamwhaley:
Is this one stuck @sboeuf @jodh-intel @mcastelino - not clear to me what we need to do to nudge this on through CI etc.

mcastelino:
Closing this and will reopen new PR. CI seems to failing for no reason.

Issue #626, opened 2017-09-25
Configuration: Add huge page support to node level configuration


mcastelino:
Enable huge pages support at a node level. This results in all
    containers on that node being launched using huge pages.

    This is useful when these nodes are used for NFV workloads
    where there will be user mode networking stacks in use on that
    host or within the container.

PR #627, opened 2017-09-25
Configuration: Huge pages support

mcastelino:
Enable huge pages support at a node level. This results in all
containers on that node being launched using huge pages.

This is useful when these nodes are used for NFV workloads
where there will be user mode networking stacks in use on that
host or within the container.

Fixes https://github.com/clearcontainers/runtime/issues/626


config.go
```diff
@@ -232,7 +233,7 @@ func newQemuHypervisorConfig(h hypervisor) (vc.HypervisorConfig, error) {
 		DefaultVCPUs:          h.defaultVCPUs(),
 		DefaultMemSz:          h.defaultMemSz(),
 		DisableBlockDeviceUse: h.DisableBlockDeviceUse,
-		MemPrealloc:           h.MemPrealloc,
```
grahamwhaley: can you expand on why the MemPrealloc change as well pls?

grahamwhaley:
just to note - CI's and QA failing no doubt due to pending re-vendoring. Maybe we should mark such PRs as WIP until we know the missing parts have landed so we don't invoke CI on things we know will fail etc.

config.go
mcastelino: @grahamwhaley no, that is a bug. I will fix it. The two are independent.

sameo:
@mcastelino Could you please rebase that one? It should pass now.

devimc:
lgtm

PR #629, opened 2017-09-26
list: Ensure "--cc-all" details are correct

### LINKED ISSUE #628: `list --cc-all` can show incorrect details
# Description of problem

`cc-runtime list --cc-all` shows extended container details but the resources it displays are not necessarily correct.

For example, if a container is running and the kernel + image assets are updated (say, by `apt-get update`-ing the system), `cc-runtime list --cc-all` should continue to show the *old* asset versions for the container started before the update occurred.

# Expected result

Running `cc-runtime list --cc-all` should show the extended details of the actual running container.

# Actual result

`cc-runtime list --cc-all` shows the extended details for the *latest* versions of the assets (incorrect).

jodh-intel:
If a container was using old assets (kernel + image) and those assets
were then updated, "list --cc-all" would show, incorrectly, that the
existing container was using the latest asset versions, not the correct
older versions.

Removed a couple of now-incorrect tests and added a new one to assert
the correct list behaviour.

Fixes #628.

### DIFF
--- a/list.go
+++ b/list.go
@@ -17,7 +17,6 @@ package main
 
 import (
 	"encoding/json"
-	"errors"
 	"fmt"
 	"os"
 	"text/tabwriter"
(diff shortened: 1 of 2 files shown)


jodh-intel:
@sameo, @grahamwhaley - another PR with a "stuck" coveralls update. The report for this PR is here:

- https://coveralls.io/jobs/29689695

sameo:
LGTM

Issue #622, opened 2017-09-25
cri-o tests fail randomly


jodh-intel:
Occasionally, the `ctr execsync` cri-o test fails:

```
++ time bats --tap crio.bats
1..16
ok 1 ctr not found correct error message
ok 2 ctr termination reason Completed
ok 3 ctr termination reason Error
ok 4 ctr remove
ok 5 ctr lifecycle
ok 6 ctr logging
ok 7 ctr logging [tty=true]
ok 8 ctr list filtering
ok 9 ctr list label filtering
ok 10 ctr metadata in list & status
ok 11 ctr execsync conflicting with conmon flags parsing
ok 12 ctr execsync
ok 13 ctr device add
not ok 14 ctr execsync failure
# (in test file crio.bats, line 546)
#   `[ "$status" -eq 0 ]' failed
```
(for example http://cc-jenkins-ci.westus2.cloudapp.azure.com/job/clear-containers-runtime-azure-ubuntu-16-04/178/consoleText)

Related: https://github.com/clearcontainers/runtime/issues/521

jodh-intel:
@grahamwhaley, @sameo, @sboeuf - I think this may be edging towards a P1 due to the impact this is having on our PRs. Thoughts?

@sboeuf - did I hear you say you were planning to look at crio this week... ? :)

grahamwhaley:
It would sure be nice to fix it. Do we have any ideas at all what makes it happen?

jodh-intel:
I have zero idea. @sboeuf ?

sboeuf:
@jodh-intel I have no idea right now either, but yes I would like to see those sporadic issues go away. This is a P1 IMO.

jodh-intel:
Hi @sboeuf - Assigning to you since presumably this is part of #676 (?)

sboeuf:
yeah sure no problem, we have a fix for that ;)

Issue #294, opened 2017-07-04
"list" doesn't show the container owner


jodh-intel:
The `list` command shows a blank owner for each container.

jodh-intel:
I think we might want to revive this one. Since we do not display an owner, that field essentially "evaporates" which is not helpful if other tools are trying to parse the output (currently 5 fields will be displayed by default per container, but if we actually displayed an owner, there would be 6).

We either need to set the owner field or maybe display `""` or `-` so that the runtime always displays the same number of fields (6 by default).

PR #631, opened 2017-09-26
scripts: Add script to find stale containers

### LINKED ISSUE #630: Add feature to find "stale" containers
If the assets are updated (kernel + rootfs image), there should be a way to determine which containers that were running at the time of the asset updated are now "stale" (really need to be restarted to take advantage of the new assets for functionality/performance/security improvements).

The simplest solution is a script, but a better way is to modify `list --cc-all` to show whether the assets used for each container are current or not.

jodh-intel:
A stale container is one which was running at the time assets (kernel +
rootfs image) were updated: they are using older asset versions than
are currently available on the system.

Provide a script to determine these stale containers to allow an
administrator to decide if they should be restarted to take advantage
of the new asset versions.

Fixes #630.

### DIFF
--- a/data/cc-find-stale-containers.sh
+++ b/data/cc-find-stale-containers.sh
@@ -1,0 +1,110 @@
+#!/bin/bash
+#
+# Copyright (c) 2017 Intel Corporation
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#      http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+set -e
+
+verbose=0
+script_name=${0##*/}
+
+usage()
+{
+    cat <<EOT
+Usage: $script_name [options]
+
+Description: Find running containers that are not using the latest assets
+             (image and kernel).
+
+Options:
+
+  -h : Show this help.
+  -v : Show containers as they are being checked.
+
+Exit codes:
+
+    0 on success.
+    1 on error.
+    2 if any containers are found to be "stale"
+      (not using the latest assets).
+
+EOT
+}
+
+die()
+{
+    msg="$*"
+    echo >&2 "ERROR: $msg"
+    exit 1
+}
+
+while getopts "hv" opt
+do
+    case "$opt" in
+        h)
+            usage
+            exit 0
+            ;;
+        v)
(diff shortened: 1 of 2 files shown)


jodh-intel:
Blocked on #629.

jodh-intel:
Remove dnm label now #629 is merged.

data/cc-find-stale-containers.sh
```diff
@@ -0,0 +1,110 @@
...
+
+    if (verbose == 1) {
+        printf("Checking container '%s'\n", container)
+    }
+
+    # XXX: See https://github.com/clearcontainers/runtime/issues/294
+    if (NF == 8) {
+        kernel_field = $7
+        image_field = $8
+    } else if (NF == 7) {
```
grahamwhaley: this looks oddly like, if we have 7 fields, use fields 8 and 9 (I was expecting maybe 6 and 7) - am I missing something?

jodh-intel: Ah - good catch. Updated.

jodh-intel:
Another occurrence of #622.

Actually, I'm glad it isn't merged yet since I've had a change of heart: rather than creating yet-another-script, how about we just extend the fields displayed by `--cc-all` to add one more that defines whether any assets are "stale"? We have full control over the fields displayed with that option as it's not a standard one. Hence, we could have something like:

- All assets for container are up-to-date:
  ```
  STALE
  no
  ```
- Single asset not current:
  ```
  STALE
  image
  ```

- List of assets for container that are stale:
  ```
  STALE
  kernel,image
  ```
@sameo, @grahamwhaley - what do you think - script or built-in feature [*]?

---
[*] - which is probably less likely to go stale itself ;-)

jodh-intel:
Added dnm label until we've finished this discussion.

grahamwhaley:
I think moving it to `cc-runtime list --cc-all` or thereabouts is probably the right move.
The output from that is already somewhat 'wide' - another column, what will it hurt ;-)
I think the column data above would be OK. I tried to think up some sort of 'inverted' version that would make it more obvious which ones were stale, but could not come up with anything. I think the 'STALE: no' item bugs me as I'd probably rather just have the column entry empty if nothing was stale - but then that puts a 'gap' in the column formatting.

jodh-intel:
I agree. I don't really like the name `STALE` for a column heading. Maybe it could be `UPDATABLE` with values of `no`, or `kernel`, `image,kernel`, or `image`?

Yes, I'd rather we don't have any (more) "gaps" as #294 is already annoying in that the number of fields can vary: imho that number of fields should always be fixed.

Another tack we couild take is to add a new option to `list`. Something like `list --cc-show-all-stale-containers` that would *only* show a subset of those containers that were actually stale.

grahamwhaley:
I considered a separate option just to show the stale containers - and I think that might be best. +1 from me.

jodh-intel:
Closing as we don't want to use a script now.

Issue #281, opened 2017-06-30
cc-check should only warn if nested kvm is not available


jodh-intel:
Nested kvm isn't strictly necessary to run a Clear Container. But without it, users won't be able to run a fully virtualised VM inside the Clear Container (which might be surprising).

Make `cc-check` warn if nested kvm support isn't available.

PR #639, opened 2017-09-27
cc-check: Always run all tests

### LINKED ISSUE #638: cc-check should run all checks and display all warnings and errors
`cc-runtime cc-check` currently runs a series of tests but crucially errors when any test fails.

This is a problem because, for example, if nesting is not enabled, currently `cc-check` will fail. However, nesting isn't strictly necessary (see #281) and since `cc-check` fails for this, it masks the latest checks that really are hard requirements (like vhost being available).

We need to change the logic of `cc-check` so that it always runs all tests and displays all results. If any error occurs, it should error only after running all tests.

jodh-intel:
Change "cc-check" so that it always runs all commands before erroring.

Previously, it would fail on first error, but this could mask the real
reason for failure. For example, on a system with nesting and vhost
disabled, "cc-check" would fail with a nesting error. But the nesting
check is not strictly required (#281) whereas vhost definitely is which
lead to confusing results for users.

"cc-check" will now run all tests and display an error message at the end
if any errors were found in the set of tests run.

Fixes #638.

### DIFF
--- a/cc-check.go
+++ b/cc-check.go
@@ -205,6 +205,11 @@ func checkKernelModules(modules map[string]kernelModule) (count uint32, err erro
 					continue
 				}
 
+				if param == "nested" {
+					ccLog.Warn(msg)
+					continue
+				}
+
 				ccLog.Error(msg)
 				count++
 			}
--- a/cc-check.go
+++ b/cc-check.go
@@ -15,7 +15,6 @@
 package main
 
 import (
-	"errors"
 	"fmt"
 	"os/exec"
 	"path/filepath"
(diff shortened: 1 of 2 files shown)


sameo:
LGTM

[Approved with PullApprove](https://pullapprove.com/clearcontainers/runtime/pull-request/639/?utm_source=github-pr&utm_medium=comment-badge&utm_campaign=clearcontainers/runtime) [Approved with PullApprove](https://pullapprove.com/clearcontainers/runtime/pull-request/639/?utm_source=github-pr&utm_medium=comment-badge&utm_campaign=clearcontainers/runtime)

jodh-intel:
Again, coverall is stuck - the report is here: https://coveralls.io/jobs/29737580 (successful).

PR #643, opened 2017-09-28
fix write cgroup no space left

### LINKED ISSUE #642: oci runtime error: no space left on device
I found below error in CC 3.0.0.

```
$sudo docker run -it --rm --runtime=runc centos pwd
/

$sudo docker run -it --rm centos bash
docker: Error response from daemon: oci runtime error: write /sys/fs/cgroup/cpu/docker/ff666f1149c9fd19ed254564b412086d1dbf2ed24d29262fac8946c4decd5a4f/tasks: no space left on device.

$sudo cc-runtime cc-env
[Meta]
  Version = "1.0.3"

[Runtime]
  [Runtime.Version]
    Semver = "3.0.0"
    Commit = "7fc11d5"
    OCI = "1.0.0-rc5"
  [Runtime.Config]
    GlobalLogPath = "/var/lib/clear-containers/runtime/runtime.log"
    [Runtime.Config.Location]
      Path = "/usr/share/defaults/clear-containers/configuration.toml"
      Resolved = "/usr/share/defaults/clear-containers/configuration.toml"

[Hypervisor]
  MachineType = "pc"
  Version = "QEMU emulator version 2.7.0, Copyright (c) 2003-2016 Fabrice Bellard and the QEMU Project developers"
  [Hypervisor.Location]
    Path = "/usr/bin/qemu-system-x86_64"
    Resolved = "/usr/bin/qemu-system-x86_64"

[Image]
  Path = "/usr/share/clear-containers/clear-containers.img"
  Resolved = "/usr/share/clear-containers/clear-17270-containers.img"

[Kernel]
  Parameters = ""
  [Kernel.Location]
    Path = "/usr/share/clear-containers/vmlinuz.container"
    Resolved = "/usr/share/clear-containers/vmlinuz-4.9.47-77.2.container"

[Proxy]
  Type = "ccProxy"
  Version = "Version: 3.0.0+git.c66f21e"
  URL = "unix:///var/run/clear-containers/proxy.sock"

[Shim]
  Type = "ccShim"
  Version = "shim version: 3.0.0-alpha.4 (commit: 739e5e2)"
  [Shim.Location]
    Path = "/usr/libexec/clear-containers/cc-shim"
    Resolved = "/usr/libexec/clear-containers/cc-shim"

[Agent]
  Type = "hyperstart"
  Version = "<<unknown>>"
  [Agent.PauseBin]
    Path = "/var/lib/clear-containers/runtime/bundles/pause_bundle/bin/pause"
    Resolved = "/var/lib/clear-containers/runtime/bundles/pause_bundle/bin/pause"

[Host]
  Kernel = "3.10.0-693.2.2.el7.x86_64"
  CCCapable = false
  [Host.Distro]
    Name = "Alibaba Group Enterprise Linux Server"
    Version = "7.2"
  [Host.CPU]
    Vendor = "GenuineIntel"
    Model = "Intel(R) Xeon(R) CPU E5-2430 0 @ 2.20GHz"
```

Ace-Tang:
fix write /sys/fs/cgroup/cpu/docker/471fb846c4d02703eebe9d3bafdbba3a5975b6c6535492784036298705576457/tasks: no space left on device

because cpuset.mems in the container's cpuset cgroup directory is empty
Fix #642 

Signed-off-by: Ace-Tang <huamin.thm@alibaba-inc.com>

### DIFF
--- a/create.go
+++ b/create.go
@@ -18,8 +18,10 @@ package main
 import (
 	"errors"
 	"fmt"
+	"io/ioutil"
 	"os"
 	"path/filepath"
+	"strings"
 
 	"github.com/Sirupsen/logrus"
 	vc "github.com/containers/virtcontainers"
@@ -134,7 +136,13 @@ func create(containerID, bundlePath, console, pidFilePath string, detach bool,
 		return err
 	}
 
-	if err := createCgroupsFiles(containerID, cgroupsPathList, process.Pid); err != nil {
+	// cgroupsDirPath is CgroupsPath fetch from OCI spec
+	var cgroupsDirPath string
+	if ociSpec.Linux != nil {
+		cgroupsDirPath = ociSpec.Linux.CgroupsPath
+	}
+
+	if err := createCgroupsFiles(containerID, cgroupsDirPath, cgroupsPathList, process.Pid); err != nil {
 		return err
 	}
 
@@ -218,7 +226,7 @@ func createContainer(ociSpec oci.CompatOCISpec, containerID, bundlePath,
 	return c.Process(), nil
 }
 
-func createCgroupsFiles(containerID string, cgroupsPathList []string, pid int) error {
+func createCgroupsFiles(containerID string, cgroupsDirPath string, cgroupsPathList []string, pid int) error {
 	if len(cgroupsPathList) == 0 {
 		fields := logrus.Fields{
 			"container": containerID,
@@ -231,6 +239,11 @@ func createCgroupsFiles(containerID string, cgroupsPathList []string, pid int) e
 	for _, cgroupsPath := range cgroupsPathList {
 		if err := os.MkdirAll(cgroupsPath, cgroupsDirMode); err != nil {
 			return err
+		}
+
+		if strings.Contains(cgroupsPath, "cpu") && cgroupsDirPath != "" {
+			parent := strings.TrimSuffix(cgroupsPath, cgroupsDirPath)
+			copyParentCPUSet(cgroupsPath, parent)
 		}
 
 		tasksFilePath := filepath.Join(cgroupsPath, cgroupsTasksFile)
@@ -288,4 +301,49 @@ func createPIDFile(pidFilePath string, pid int) error {
 	}
 
 	return nil
+}
+
+// copyParentCPUSet copies the cpuset.cpus and cpuset.mems from the parent
+// directory to the current directory if the file's contents are 0
+func copyParentCPUSet(current, parent string) error {
+	currentCpus, currentMems, err := getCPUSet(current)
+	if err != nil {
+		return err
+	}
+
+	parentCpus, parentMems, err := getCPUSet(parent)
+	if err != nil {
+		return err
+	}
+
+	if len(parentCpus) < 1 || len(parentMems) < 1 {
+		return nil
+	}
+
+	var cgroupsFileMode = os.FileMode(0600)
+	if isEmptyString(currentCpus) {
+		if err := writeFile(filepath.Join(current, "cpuset.cpus"), string(parentCpus), cgroupsFileMode); err != nil {
+			return err
+		}
+	}
+
+	if isEmptyString(currentMems) {
+		if err := writeFile(filepath.Join(current, "cpuset.mems"), string(parentMems), cgroupsFileMode); err != nil {
+			return err
+		}
+	}
+
+	return nil
+}
+
+func getCPUSet(parent string) (cpus []byte, mems []byte, err error) {
+	if cpus, err = ioutil.ReadFile(filepath.Join(parent, "cpuset.cpus")); err != nil {
+		return
+	}
+
+	if mems, err = ioutil.ReadFile(filepath.Join(parent, "cpuset.mems")); err != nil {
+		return
+	}
+
+	return cpus, mems, nil
 }
--- a/create_test.go
+++ b/create_test.go
@@ -42,14 +42,14 @@ const (
 
 var testStrPID = fmt.Sprintf("%d", testPID)
 
-func testCreateCgroupsFilesSuccessful(t *testing.T, cgroupsPathList []string, pid int) {
-	if err := createCgroupsFiles("foo", cgroupsPathList, pid); err != nil {
+func testCreateCgroupsFilesSuccessful(t *testing.T, cgroupsDirPath string, cgroupsPathList []string, pid int) {
+	if err := createCgroupsFiles("foo", cgroupsDirPath, cgroupsPathList, pid); err != nil {
 		t.Fatalf("This test should succeed (cgroupsPath %q, pid %d): %s", cgroupsPathList, pid, err)
 	}
 }
 
 func TestCgroupsFilesEmptyCgroupsPathSuccessful(t *testing.T) {
-	testCreateCgroupsFilesSuccessful(t, []string{}, testPID)
+	testCreateCgroupsFilesSuccessful(t, "", []string{}, testPID)
 }
 
 func TestCreateCgroupsFilesFailToWriteFile(t *testing.T) {
@@ -71,7 +71,7 @@ func TestCreateCgroupsFilesFailToWriteFile(t *testing.T) {
 
 	files := []string{file}
 
-	err = createCgroupsFiles("foo", files, testPID)
+	err = createCgroupsFiles("foo", "cgroups-file", files, testPID)
 	assert.Error(err)
 }
 
@@ -81,7 +81,7 @@ func TestCgroupsFilesNonEmptyCgroupsPathSuccessful(t *testing.T) {
 		t.Fatalf("Could not create temporary cgroups directory: %s", err)
 	}
 
-	testCreateCgroupsFilesSuccessful(t, []string{cgroupsPath}, testPID)
+	testCreateCgroupsFilesSuccessful(t, "cgroups-path-", []string{cgroupsPath}, testPID)
 
 	defer os.RemoveAll(cgroupsPath)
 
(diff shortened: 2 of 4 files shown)


jovizhangwei:
Blocked by this `no space left on device` issue, hopefully it can be merged ASAP.

jodh-intel:
Hi @Ace-Tang - thanks for raising this. Before we can consider this there are a few things you need to do:

- Reformat your commit slightly:
  - Raise and issue and add a "Fixes #XXX2 in the commit.
  - Add a subsystem (`something: ...` prefix) in your commit.
   See: https://github.com/clearcontainers/runtime/pull/619#issuecomment-331874714
- Write tests for the new functions you have added.

Also, the code will need to be reviewed by a couple of team members. I'll take a look now...

create.go
```diff
@@ -284,3 +292,58 @@ func createPIDFile(pidFilePath string, pid int) error {
...
+	return len(bytes.Trim(b, "\n")) == 0
+}
+
+func writeFile(dir, file, data string) error {
+	// Normally dir should not be empty, one case is that cgroup subsystem
+	// is not mounted, we will get empty dir, and we want it fail here.
+	if dir == "" {
+		return fmt.Errorf("no such directory for %s", file)
+	}
+	if err := ioutil.WriteFile(filepath.Join(dir, file), []byte(data), 0700); err != nil {
```
jodh-intel: Rather than `0700` this should probably be `cgroupsFileMode` (but see comment above).

create.go
```diff
@@ -217,7 +220,7 @@ func createContainer(ociSpec oci.CompatOCISpec, containerID, bundlePath,
 	return c.Process(), nil
 }
 
-func createCgroupsFiles(cgroupsPathList []string, pid int) error {
+func createCgroupsFiles(cgParentPath string, cgroupsPathList []string, pid int) error {
```
jodh-intel: What if `cgParentPath` is `""`? That value is optional:

- https://github.com/opencontainers/runtime-spec/blob/master/config-linux.md#cgroups-path

create.go
```diff
@@ -284,3 +292,58 @@ func createPIDFile(pidFilePath string, pid int) error {
...
+	}
+	if isEmpty(currentMems) {
+		if err := writeFile(current, "cpuset.mems", string(parentMems)); err != nil {
+			return err
+		}
+	}
+	return nil
+}
+
+func getSubsystemSettings(parent string) (cpus []byte, mems []byte, err error) {
```
jodh-intel: Missing function header to explain what this does. Also, I'd prefer the name to container `cgroup` to make it clearer what it is operating on.

create.go
```diff
@@ -284,3 +292,58 @@ func createPIDFile(pidFilePath string, pid int) error {
 
 	return nil
 }
+
+// copyIfNeeded copies the cpuset.cpus and cpuset.mems from the parent
+// directory to the current directory if the file's contents are 0
+func copyIfNeeded(current, parent string) error {
```
jodh-intel: Can you add `cgroup` to the name of this function for clarity?

create.go
```diff
@@ -284,3 +292,58 @@ func createPIDFile(pidFilePath string, pid int) error {
...
+	var (
+		err                      error
+		currentCpus, currentMems []byte
+		parentCpus, parentMems   []byte
+	)
+
+	if currentCpus, currentMems, err = getSubsystemSettings(current); err != nil {
+		return err
+	}
+	if parentCpus, parentMems, err = getSubsystemSettings(parent); err != nil {
```
jodh-intel: Missing blank line before `if...` (this comment also applies in other places in this PR).

create.go
```diff
@@ -284,3 +292,58 @@ func createPIDFile(pidFilePath string, pid int) error {
...
+		return
+	}
+	return cpus, mems, nil
+}
+
+func isEmpty(b []byte) bool {
+	return len(bytes.Trim(b, "\n")) == 0
+}
+
+func writeFile(dir, file, data string) error {
```
jodh-intel: This is a relatively general-purpose routine so I would either move it to `utils.go` and change it to be:

```
func writefile(filePath, data string) error
```

.. or make maybe make it a cgroup-specific function by changing its name.

create.go
sameo: I would just call that one `copyParentCpuSet()`
The `IfNeeded` part is not necessary.

create.go
sameo: Here as well, the naming is too generic. What your doing is getting cpuset values, so please rename to:

```Go
getCpuSet(cgCpuset string) (cpus []byte, mems []byte, err error)
```

create.go
sameo: if `cgParentPath` is optional, we should fallback to `cgroupsDirPath`.

Ace-Tang:
@sameo  @jodh-intel     thanks for review my code and give me  advisor, I will  modify the code

create.go
Ace-Tang: I wonder if  `isEmpty` should be moved to util.go
```
func isEmpty(b []byte) bool
```

sameo: @Ace-Tang Yes, good point. Let's move it there, and also rename it to `isEmptyString()`.

create.go
```diff
@@ -284,3 +292,58 @@ func createPIDFile(pidFilePath string, pid int) error {
...
 }
+
+// copyIfNeeded copies the cpuset.cpus and cpuset.mems from the parent
+// directory to the current directory if the file's contents are 0
+func copyIfNeeded(current, parent string) error {
+	var (
+		err                      error
+		currentCpus, currentMems []byte
+		parentCpus, parentMems   []byte
+	)
```
sboeuf: No need to declare those variables here.
Please do the following function call like this:
```go
currentCpus, currentMems, err := getSubsystemSettings(current)
if err != nil {
        return err
}

parentCpus, parentMems, err := getSubsystemSettings(parent)
if err != nil {
        return err
}
```

Ace-Tang:
@sboeuf  please help check my commit message, thanks

@houndci-bot  thanks for the suggestion about my function name

sameo:
LGTM

sameo:
@Ace-Tang Thanks a lot for the fix and the quick turnaround.

create_test.go
```diff
@@ -1059,3 +1059,50 @@ func TestCreateCreateContainer(t *testing.T) {
...
+	}
+
+	assert := assert.New(t)
+
+	cgroupsRootPath := findCgroupRoot()
+	if cgroupsRootPath == "" {
+		t.Skip(testDisabledNeedCgroup)
+	}
+
+	cgroupsCPUSetRoot := filepath.Join(cgroupsRootPath, "cpu")
```
sameo: This should be:

```Go
cgroupsCPUSetRoot := filepath.Join(cgroupsRootPath, "cpuset")
```

for the test to pass (as sudo).

sameo:
@Ace-Tang I copied your PR into #649 because we changed our CI and I wanted to check if it passes with the new CI jobs (and potentially merge yours if #649 would have passed). Unfortunately it does not, because I think one of your test is wrong. I commented on the code itself, but essentially the fix should be:

```sh
diff --git a/create_test.go b/create_test.go
index 6e29d79..7c80b26 100644
--- a/create_test.go
+++ b/create_test.go
@@ -1086,7 +1086,7 @@ func TestCopyParentCPUSetSuccessful(t *testing.T) {
                t.Skip(testDisabledNeedCgroup)
        }
 
-       cgroupsCPUSetRoot := filepath.Join(cgroupsRootPath, "cpu")
+       cgroupsCPUSetRoot := filepath.Join(cgroupsRootPath, "cpuset")
        if _, err := os.Stat(cgroupsCPUSetRoot); err != nil && os.IsNotExist(err) {
                t.Skip(testDisabledNeedCgroup)
        }

```

Could you please fix the test and resubmit ? Thanks in advance.
I'm closing #649 now.

create_test.go
Ace-Tang: thanks for testing my code.  the test is using for testing copy data from /sys/fs/cgroup/cpu/cpuset.cpus and /sys/fs/cgroup/cpu/cpuset.mems to container's cgroup path, I test this on my redhat machine many times,  it can pass always, and now I check this on my ubuntu machine , it shows that 
```
root@www:/home/ace/gosrc/src/github.com/clearcontainers/runtime# ls /sys/fs/cgroup/cpu
cgroup.clone_children  cgroup.sane_behavior  cpuacct.usage         cpu.cfs_period_us  cpu.shares  docker      notify_on_release  system.slice  user.slice
cgroup.procs           cpuacct.stat          cpuacct.usage_percpu  cpu.cfs_quota_us   cpu.stat    init.scope  release_agent      tasks

```
looks like cgroup system on ubuntu and redhat is different, ubuntu not have cpuset.cpus and couset.mem in cpu cgroup path as redhat does
I really sorry for not test this on ubuntu machine, I will update my test

do I need to reopen a PR, or still use #643   @sameo

Ace-Tang: I add more check for the test
```
diff --git a/create_test.go b/create_test.go
index 6e29d79..6089047 100644
--- a/create_test.go
+++ b/create_test.go
@@ -1091,6 +1091,16 @@ func TestCopyParentCPUSetSuccessful(t *testing.T) {
                t.Skip(testDisabledNeedCgroup)
        }
 
+       cpusetCpusPath := filepath.Join(cgroupsCPUSetRoot, "cpuset.cpus")
+       if _, err := os.Stat(cpusetCpusPath); err != nil && os.IsNotExist(err) {
+               t.Skip(testDisabledNeedCgroup)
+       }
+
+       cpusetMemsPath := filepath.Join(cgroupsCPUSetRoot, "cpuset.mems")
+       if _, err := os.Stat(cpusetMemsPath); err != nil && os.IsNotExist(err) {
+               t.Skip(testDisabledNeedCgroup)
+       }
+

```

sameo: @Ace-Tang On both Ubuntu and Fedora (26) the cpu cgroup does not have a cpuset. I believe any distro with a fairly recent kernel (4.x) will have that behaviour. You see an empty cpuset on the cpu cgroup on older kernels (RHEL, Centos).

create_test.go
```diff
@@ -1059,3 +1059,60 @@ func TestCreateCreateContainer(t *testing.T) {
...
+	cpusetMemsPath := filepath.Join(cgroupsCPUSetRoot, "cpuset.mems")
+	if _, err := os.Stat(cpusetMemsPath); err != nil && os.IsNotExist(err) {
+		t.Skip(testDisabledNeedCgroup)
+	}
+
+	cgroupsPath, err := ioutil.TempDir(cgroupsCPUSetRoot, "cgroups-path-")
+	if err != nil {
+		t.Fatalf("Could not create temporary cgroups directory: %s", err)
+	}
+	defer os.RemoveAll(cgroupsPath)
```
sameo: That's good, it won't fail on distros where cpuset is not created under the cpu cgroup.
However, since we're running our CI on Ubuntu and FC26, it will prevent us from testing the positive path of `copyParentCPUSet()` on our CI (The above tests will always fail and thus fall into skip). Could you please add another test where we run exactly the same code path, except we do it under `/sys/fs/cgroup/cpuset`, so that we unit test both cases ? This does not have to be part of this PR (provided the code coverage is fine), it can be in a next one after we merge that one (I want to merge it asap to get you guys rolling).

PR #654, opened 2017-09-29
tests: Only pull docker test image if not available locally

### LINKED ISSUE #653: only pull test docker image if not present locally
The unit-tests need a `busybox` docker image to generate an OCI bundle from. Currently, the tests always `docker pull busybox` *twice* (once for a non-priv user and once for `root`).

Modify the behaviour so that if the image exists locally, don't call `docker pull busybox` at all, which will speed up the overall unit-test runtime.

jodh-intel:
Change the unit-test setup code to only download the "busybox" docker
image if it doesn't already exist locally. Since the unit-tests are run
twice (once as a non-privileged user and once as `root`), this can speed
up the overall unit-test runtime significantly.

Fixes #653.

### DIFF
--- a/main_test.go
+++ b/main_test.go
@@ -104,9 +104,16 @@ func init() {
 	// Do this now to avoid hitting the test timeout value due to
 	// slow network response.
 	fmt.Printf("INFO: ensuring required docker image (%v) is available\n", testDockerImage)
-	_, err = runCommand([]string{"docker", "pull", testDockerImage})
-	if err != nil {
-		panic(err)
+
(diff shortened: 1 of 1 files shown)


jodh-intel:
Can we get a quick review on this trivial change?

jodh-intel:
Well, difficult to give accurate figures for all cases of course as this depends on network speeds.

But if you're interested, a `docker pull busybox` can take 4 seconds for me. An image inspect is fast so overall this now take a few centi-seconds for me rather than ~8 seconds.

PR #659, opened 2017-10-01
tests: switch to test copyParentCPUSet under tmpDir

Ace-Tang:
test copyParentCPUSet() under tmpDir instead of using real
cgroups syatem path, make the test always worked whether
cpuset.cpus and cpuset.mems files in cpu cgroups exist or not.

Signed-off-by: Ace-Tang <aceapril@126.com>

please help me check if the patch is properly, thanks.
ping @sameo

### DIFF
--- a/create_test.go
+++ b/create_test.go
@@ -42,6 +42,16 @@ const (
 
 var testStrPID = fmt.Sprintf("%d", testPID)
 
+func mockCPUSetContent(contents map[string]string) error {
+	for filePath, data := range contents {
+		if err := writeFile(filePath, data, testFileMode); err != nil {
+			return err
+		}
+	}
+
+	return nil
+}
+
 func testCreateCgroupsFilesSuccessful(t *testing.T, cgroupsDirPath string, cgroupsPathList []string, pid int) {
 	if err := createCgroupsFiles("foo", cgroupsDirPath, cgroupsPathList, pid); err != nil {
 		t.Fatalf("This test should succeed (cgroupsPath %q, pid %d): %s", cgroupsPathList, pid, err)
@@ -1064,9 +1074,7 @@ func TestCopyParentCPUSetFail(t *testing.T) {
 	assert := assert.New(t)
 
 	cgroupsPath, err := ioutil.TempDir(testDir, "cgroups-path-")
-	if err != nil {
-		t.Fatalf("Could not create temporary cgroups directory: %s", err)
-	}
+	assert.NoError(err)
 	defer os.RemoveAll(cgroupsPath)
 
 	err = copyParentCPUSet(cgroupsPath, testDir)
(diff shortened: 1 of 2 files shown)


sameo:
@Ace-Tang The patch looks good to me, but you need to create an issue for your PR and then add a `Fixes` line to your commit message.

Ace-Tang:
oh, sorry, I do not konw issue is necessary， I will update soon

create_test.go
```diff
@@ -42,6 +42,17 @@ const (
 
 var testStrPID = fmt.Sprintf("%d", testPID)
 
+func mockCPUSetContent(contents map[string]string) error {
+	fileMode := os.FileMode(0600)
```
jodh-intel: You could just use `testFileMode` (from `main_test.go`) here.

create_test.go
```diff
@@ -1074,43 +1085,50 @@ func TestCopyParentCPUSetFail(t *testing.T) {
...
-		t.Skip(testDisabledNeedRoot)
-	}
-
 	assert := assert.New(t)
 
-	cgroupsRootPath := findCgroupRoot()
-	if cgroupsRootPath == "" {
-		t.Skip(testDisabledNeedCgroup)
+	cgroupsPath, err := ioutil.TempDir(testDir, "cgroups-path-")
+	if err != nil {
```
jodh-intel: All the `if err!= nil` tests can be replaced with the simpler and shorter `assert.NoErrror()` syntax. For example:

```
assert.NoErrror(err, "Could not create temporary cgroups directory")
```

create_test.go
Ace-Tang: yes, I also consider whether to use testFileMode or use `os.FileMode(0600)` here

create_test.go
Ace-Tang: I follow the style from this function
```
func TestCgroupsFilesNonEmptyCgroupsPathSuccessful(t *testing.T) {
	cgroupsPath, err := ioutil.TempDir(testDir, "cgroups-path-")
	if err != nil {
		t.Fatalf("Could not create temporary cgroups directory: %s", err)
	}
       ...
```

and I found that this code style exist in another function `TestCreatePIDFileSuccessful`
```
func TestCreatePIDFileSuccessful(t *testing.T) {
      pidDirPath, err := ioutil.TempDir(testDir, "pid-path-")
      if err != nil {
          t.Fatalf("Could not create temporary PID directory: %s", err)
      }
 ... 

```
I wonder if we can use `assert.NoError(err)` to simplify code for these functions, and make style  consistent

jodh-intel: Yes, there is still test code that uses the old `if err != nil` style, but we are trying to make all new test code use `assert.*` as it is simpler, easier to read and shorter ;) When we get a chance, we'll update the older test code to use `assert.*` too.

As I say, whichever style you chose isn't going to block this code landing, but it's just an "FYI" for future.

Ace-Tang: @jodh-intel  , I update the code, use `assert.*` instead of `if err != nil`

jodh-intel:
Thanks @Ace-Tang!

@sboeuf, @grahamwhaley - can you restart the 16.04 build for this PR? I cannot as I see a bunch of errors like this:
```
Error while checking in file to scm repository
```

... and they are covering up the "restart build" button! If I try to remove the message, I get an error that only an admin can purge log messages.

grahamwhaley:
@jodh-intel Oh, I see what you mean by 'covering up' the button - there was a not-easy-to-see 'clear this error' link, so I pressed that, which then cleared it down and allowed me to hit the rebuild - should be in progress now

jodh-intel:
@grahamwhaley - thanks - that link is what threw the Jenkins perm error at me (I don't have your super-powers! ;)

sboeuf:
@jodh-intel I have added you now, you should be able to rebuild from Jenkins

sameo:
LGTM

sameo:
Coveralls says coverage is increasing with that one: https://coveralls.io/builds/13561840

PR #670, opened 2017-10-04
list: Add columns to extended output to show stale assets

### LINKED ISSUE #630: Add feature to find "stale" containers

jodh-intel:
Add three new columns to the "list --cc-all" output:

 - LATEST-KERNEL (Newest locally-available kernel)
 - LATEST-IMAGE  (Newest locally-available image)
 - STALE         (String describing if the current kernel or image are not current).
Fixes #630.

### DIFF
--- a/list.go
+++ b/list.go
@@ -17,8 +17,10 @@ package main
 
 import (
 	"encoding/json"
+	"errors"
 	"fmt"
 	"os"
+	"strings"
 	"text/tabwriter"
 	"time"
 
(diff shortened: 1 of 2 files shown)


jodh-intel:
Quick! Someone review before Jenkins notices! ;)

jodh-intel:
I've created a wiki page describing stale containers and how to deal with them:

- https://github.com/clearcontainers/runtime/wiki/Stale-containers

PR #693, opened 2017-10-09
Make debug configurable

jodh-intel:
Add config options to toggle debug for the shim and the runtime (we already have an option for the hypervisor and the proxy is handled separately).

### DIFF
--- a/README.md
+++ b/README.md
@@ -69,6 +69,8 @@ $ cc-runtime cc-env
 
 ## Debugging
 
+### Global logfile
+
 To provide a persistent log of all container activity on the system, the runtime
 offers a global logging facility. By default, this feature is disabled
 but can be enabled with a simple change to the [configuration](#Configuration) file.
@@ -101,6 +103,14 @@ attempt to create it.
 
 It is the Administrator's responsibility to ensure there is sufficient
 space for the global log.
+
+### Enabling debug for various components
+
+The runtime, the shim (`cc-shim`), and the hypervisor all have separate debug
+options in the [configuration file](#Configuration).
+
+The proxy (`cc-proxy`) has a command-line option to enable debug output. See
+the [proxy documentation](https://github.com/clearcontainers/proxy#debugging) for further details.
 
 ## Limitations
 
--- a/README.md
+++ b/README.md
@@ -6,6 +6,19 @@
 
 # runtime
 
+* [Introduction](#introduction)
+* [License](#license)
+* [Hardware requirements](#hardware-requirements)
+* [Quick start for users](#quick-start-for-users)
+* [Quick start for developers](#quick-start-for-developers)
+* [Community](#community)
+* [Configuration](#configuration)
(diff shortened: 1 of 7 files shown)


jodh-intel:
Hi @iphutch - please can you review the doc changes on this PR?

mcastelino:
@jodh-intel @jcvenegas with this PR we should be able to have both debug logs from proxy as well as the ability to console into the VM right? That closes a gap in our debug environment where we could have one or the other.

jodh-intel:
Ah - the unit-tests are happy, but some of the tests in the `tests` repo aren't since the `--debug` flag has been zapped.

We could retain `--debug` as an override for whatever is specified in the config file: if `--debug` is set/cleared, enable/disable debug for all components that can have debug enabled via the config file. That would actually be a change to the current system though since the `--debug` flag currently only affects the runtime log level.

There is also the issue of the `DEFENABLEDEBUG` build variable (although that is currently broken - see https://github.com/clearcontainers/runtime/pull/590/files#r143512863).

@mcastelino - can you clarify the scope of `DEFENABLEDEBUG`? I'm not sure we want such a variable to apply to all debug options (although admittedly, the admin can disable them via the config file if they desire).

jodh-intel:
Hi @mcastelino - I'm not 100% clear on what you're saying. This PR doesn't provide anything wrt proxy debug (as that is currently only settable via a cli option (although it would be great if it could also be toggled via an API call so the runtime could control debug logs on a per pod basis).

I haven't tested this branch with your debug console one.

README.md
```diff
@@ -102,6 +117,14 @@ attempt to create it.
 It is the Administrator's responsibility to ensure there is sufficient
 space for the global log.
 
+### Enabling debug for various components
+
+The runtime, the shim (`cc-shim`), and the hypervisor all have separate debug
+options in the configuration file (see the [configuration](#Configuration) file).
```
iphutch: This reads as a little bit repetitive. Can we instead have the first "configuration" file as the link as it is on line 89?

jodh-intel: Good catch - fixed.

grahamwhaley:
@jodh-intel I think if you were to have both the config file and command line ability to configure the debug, then you're going to need `--no-debug` or equivalent adding to the command line as well I think (and of course the matching tests).
It might be easier to just drop the command line and move to the config file. Only question I think would be if any developers would miss the simple ability to run items by hand with debug enabled on the command line, vs tweaking a line in the relevant config files. I'm fine with just config files I think...

jodh-intel:
Thanks @grahamwhaley - agreed - dropping `--debug` entirely is my preference. I'll raise a PR to tweak the `tests` repo for this...

jodh-intel:
Blocked on https://github.com/clearcontainers/tests/pull/601.

jodh-intel:
@grahamwhaley - could you take a look please?

sameo:
LGTM

PR #720, opened 2017-10-13
oci: do not use specified cgroup dir path

Ace-Tang:
get cgroup directory path from mountinfo instead of
using /sys/fs/cgroup/, since maybe not all os use that path.

Fixes 719.

Signed-off-by: Ace Tang <aceapril@126.com>

### DIFF
--- a/oci.go
+++ b/oci.go
@@ -15,11 +15,13 @@
 package main
 
 import (
+	"bufio"
 	"errors"
 	"fmt"
 	"net"
 	"os"
 	"path/filepath"
+	"strings"
 	"syscall"
 
 	vc "github.com/containers/virtcontainers"
@@ -44,7 +46,9 @@ const (
 
 var errNeedLinuxResource = errors.New("Linux resource cannot be empty")
 
-var cgroupsDirPath = "/sys/fs/cgroup"
+var cgroupsDirPath string
+
+var procMountInfo = "/proc/self/mountinfo"
 
 // getContainerInfo returns the container status and its pod ID.
 // It internally expands the container ID from the prefix provided.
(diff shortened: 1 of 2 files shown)


oci.go
```diff
@@ -302,3 +310,38 @@ func noNeedForOutput(detach bool, tty bool) bool {
...
+		return "", err
+	}
+	defer f.Close()
+
+	var cgroupRootPath string
+	scanner := bufio.NewScanner(f)
+	for scanner.Scan() {
+		text := scanner.Text()
+		fields := strings.Split(text, " ")
+		index := strings.Index(text, " - ")
```
jodh-intel: Can you add a check on this ("`if index < 0`"). I'd also move the `fields :=` to the line after this one to avoid the `Split()` in case the `Index()` fails.

oci.go
```diff
@@ -302,3 +310,38 @@ func noNeedForOutput(detach bool, tty bool) bool {
...
+	if err != nil {
+		return "", err
+	}
+	defer f.Close()
+
+	var cgroupRootPath string
+	scanner := bufio.NewScanner(f)
+	for scanner.Scan() {
+		text := scanner.Text()
+		fields := strings.Split(text, " ")
```
jodh-intel: It would be a good idea to check `len(fields)` so you know that `fields[4]` below is valid.

oci.go
```diff
@@ -302,3 +310,38 @@ func noNeedForOutput(detach bool, tty bool) bool {
 
 	return true
 }
+
+func getCgroupsDirPath() (string, error) {
+	if cgroupsDirPath != "" {
+		return cgroupsDirPath, nil
+	}
+
+	f, err := os.Open("/proc/self/mountinfo")
```
jodh-intel: Can you create a global `var` for this file and add a test for this function? By using a `var` you can write a test that changes the value of it to a temp file. See for example what we currently do for `/proc/cpuinfo`:

- https://github.com/clearcontainers/runtime/blob/master/cc-check.go#L88..L103

And here's how we test it (we check what happens when the file doesn't exist, when it contains invalid data and when it contains valid data):

- https://github.com/clearcontainers/runtime/blob/master/cc-check_test.go#L47..L94

grahamwhaley:
The CI's are failing as I think you need a '#' before the Issue number on commit Fixes line, such as:
```
 Fixes #719
```

oci.go
Ace-Tang: sure, I will read it

Ace-Tang:
I update,  @jodh-intel  @grahamwhaley    PTAL :)

jodh-intel:
Coveralls run but isn't updating here. Merging...

PR #736, opened 2017-10-18
oci: Add conatiner ID to error message

### LINKED ISSUE #735: Enrich container error message
If `getExistingContainerInfo()` fails currently, it can display the error, `Container ID does not exist`. However, that error is not strictly correct and it could show the container ID that is being queried.

jodh-intel:
Improve one of the errors returned by `getExistingContainerInfo()` when
it fails to obtain the container status by including the container ID
in the returned error.

Fixes #735.

### DIFF
--- a/oci.go
+++ b/oci.go
@@ -85,7 +85,7 @@ func getExistingContainerInfo(containerID string) (vc.ContainerStatus, string, e
 
 	// container ID MUST exist.
 	if cStatus.ID == "" {
-		return vc.ContainerStatus{}, "", fmt.Errorf("Container ID does not exist")
+		return vc.ContainerStatus{}, "", fmt.Errorf("Container ID (%v) does not exist", containerID)
 	}
 
 	return cStatus, podID, nil


oci.go
```diff
@@ -85,7 +85,7 @@ func getExistingContainerInfo(containerID string) (vc.ContainerStatus, string, e
 
 	// container ID MUST exist.
 	if cStatus.ID == "" {
-		return vc.ContainerStatus{}, "", fmt.Errorf("Container ID does not exist")
+		return vc.ContainerStatus{}, "", fmt.Errorf("Invalid container status for container %v", containerID)
```
grahamwhaley: Adding the ID into the string looks good - helpful when debugging.
I'm not sure 'Invalid container status' is the correct string though. Looking at the `getContainerInfo` call just above here, I think you can only get to this place with one of two results:
1) the cStatusID will match the ID you passed in - as that is how the search does the compare/match to return your ite
1) The container with that ID does not exist (see below) - in which case, the status is not `invalid` - it is indicating there is no such container... I think `Invalid` could be misleading here.


From the tail of `getContainerInfo`:
```
// Not finding a container should not trigger an error as
	// getContainerInfo is used for checking the existence and
	// the absence of a container ID.
	return vc.ContainerStatus{}, "", nil
```

I think I'd be happier with something like:
```
fmt.Errorf("Container ID (%v) does not exist", containerID)
```

jodh-intel: Agreed - branch updated.

sameo:
LGTM

jodh-intel:
Coverall report: https://coveralls.io/jobs/30354688.

Merging...

PR #738, opened 2017-10-18
Collect script more patterns

jodh-intel:


jodh-intel:
Non-code change build so ignoring (single) CI failure.

PR #760, opened 2017-10-25
Unit test coverage fixes

### LINKED ISSUE #759: Increase unit-test coverage (again)
Due to all the problems we've seen with coveralls recently (see https://github.com/clearcontainers/jenkins/issues/9), the unit test coverage level has slipped by 3%, so fix it.

jodh-intel:
tests: Increase unit test coverage
    
Add new tests to improve the overall unit-test coverage.
    
Fixes #759.

### DIFF
--- a/cc-env_test.go
+++ b/cc-env_test.go
@@ -432,6 +432,68 @@ func TestCCEnvGetEnvInfo(t *testing.T) {
 	assert.Equal(t, expectedCCEnv, ccEnv)
 }
 
+func TestCCEnvGetEnvInfoNoHypervisorVersion(t *testing.T) {
+	assert := assert.New(t)
+
+	tmpdir, err := ioutil.TempDir("", "")
+	assert.NoError(err)
+	defer os.RemoveAll(tmpdir)
+
+	const logFile = "/tmp/file.log"
+
+	configFile, config, err := makeRuntimeConfig(tmpdir)
+	assert.NoError(err)
+
+	expectedCCEnv, err := getExpectedSettings(config, tmpdir, configFile, logFile)
+	assert.NoError(err)
+
+	err = os.Remove(config.HypervisorConfig.HypervisorPath)
+	assert.NoError(err)
+
+	expectedCCEnv.Hypervisor.Version = unknown
+
+	ccEnv, err := getEnvInfo(configFile, logFile, config)
+	assert.NoError(err)
+
+	assert.Equal(expectedCCEnv, ccEnv)
+}
+
+func TestCCEnvGetEnvInfoShimError(t *testing.T) {
+	assert := assert.New(t)
+
+	tmpdir, err := ioutil.TempDir("", "")
+	assert.NoError(err)
+	defer os.RemoveAll(tmpdir)
+
+	const logFile = "/tmp/file.log"
+
+	configFile, config, err := makeRuntimeConfig(tmpdir)
+	assert.NoError(err)
+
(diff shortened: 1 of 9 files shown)


sboeuf:
LGTM

jodh-intel:
@grahamwhaley - could you take a squizz at this PR?

Coveralls report: https://coveralls.io/jobs/30598315

config.go
```diff
@@ -226,12 +226,6 @@ func newQemuHypervisorConfig(h hypervisor) (vc.HypervisorConfig, error) {
 	kernelParams := h.kernelParams()
 	machineType := h.machineType()
 
-	for _, file := range []string{hypervisor, kernel, image} {
```
grahamwhaley: presume you found this as the coverage said it was not covered, and you could not make it so as the upper layers stopped you getting this far :-)

cc-env_test.go
```diff
@@ -432,6 +432,74 @@ func TestCCEnvGetEnvInfo(t *testing.T) {
...
+func TestCCEnvGetEnvInfoNoHypervisorVersion(t *testing.T) {
+	assert := assert.New(t)
+
+	tmpdir, err := ioutil.TempDir("", "")
+	if err != nil {
+		panic(err)
+	}
+	defer os.RemoveAll(tmpdir)
+
+	const logFile = "/tmp/file.log"
```
grahamwhaley: I see a few of these, and it feels like maybe this is wired somewhere else as well - is this something we can lift out to a single place? (but tbh, I fully expect you would have done that if it were possible).

cc-env_test.go
```diff
@@ -855,6 +923,72 @@ func TestCCEnvHandleSettings(t *testing.T) {
 	assert.NoError(t, err)
 }
 
+func TestCCEnvHandleSettingsInvalidShimConfig(t *testing.T) {
+	assert := assert.New(t)
+
+	tmpdir, err := ioutil.TempDir("", "")
+	if err != nil {
+		panic(err)
```
grahamwhaley: `panic` feels a bit harsh - OK, I know this is like a moderately fatal system error, and not strictly a test failure, but I wonder if a `Fatal` from the test suite might be more appropriate? I defer to your knowledge on that though.

delete_test.go
```diff
@@ -501,3 +501,121 @@ func TestDeleteCLIFunction(t *testing.T) {
...
+	}
+
+	pod.MockContainers = []*vcMock.Container{
+		{
+			MockID:  testContainerID,
+			MockPod: pod,
+		},
+	}
+
+	/*
```
grahamwhaley: commented out block deliberate?

grahamwhaley:
and just confirm, I believe coveralls is reporting a 94->97% bump

cc-env_test.go
jodh-intel: We probably could get rid of some of those. I do think we could do with another test sprint since although our coverage level is high, there is still quite a bit of refactoring we could do tbh.

cc-env_test.go
jodh-intel: Fixed - it was a cut'n'paste from other (older) tests that still do call `panic()` (see sprint idea above :)

config.go
jodh-intel: yes, `go test` discovered this was redundant code.

delete_test.go
jodh-intel: Oops. fixed.

Issue #763, opened 2017-10-25
Consider making `make install` configure the container manager


jodh-intel:
As pointed out by @mcastelino (https://github.com/clearcontainers/tests/pull/639/files#r146899594), we could consider having `make install` configure the runtime the container manager will use.

Questions:

- Should it create `/etc/docker/daemon.json` only if not present?
- Should it update `/etc/docker/daemon.json` if present?
- Should the default be to configure automatically (detect if `docker`/k8s is installed)?
  Probably best to only attempt to configure if `CC_CONTAINER_MANAGER=[docker|kubernetes]` and `CC_CONFIGURE_CONTAINER_MANAGER=yes` are set.
- Should we instead create `/usr/share/clear-containers/docker-daemon.json`.
  This would allow the admin to modify `docker.service` to specify `dockerd --config-file=/usr/share/clear-containers/docker-daemon.json`.
- What about k8s?

We could also enable the container managers debug option(s) if `DEFENABLEDEBUG` is set.

This all has a packaging impact so we need input from @gorozco1, @erick0z.

jodh-intel:
It's worth mentioning that this *might* actually make the configuration safer given that currently a systemd snippet is used which basically masks the package-provided set of daemon options. By only specifying the runtime (and maybe enabling debug), users can again use the vanilla packaged systemd docker unit.

Issue #754, opened 2017-10-23
Documentation/Scripts: Use daemon.json to add cc-runtime to dockerd


mcastelino:
# Description of problem

Today we recommend the use of a systemd unit file to add cc-runtime to docker. However this can cause issues if the user does not ensure that the new unit file created differs from what the distribution has setup as the default. 

Hence the safer option daemon.json to add cc-runtime to docker.

# Expected result

An example of such a daemon.json is

```
{
       "default-runtime": "cc-runtime",
        "runtimes": {
             "cc-runtime": {
                      "path": "/usr/bin/cc-runtime"
             }
        }
}

```

We have this in our documentation  in various places

https://github.com/clearcontainers/runtime/blob/master/installation/centos-setup.sh#L133
https://github.com/clearcontainers/runtime/blob/master/docs/fedora-installation-guide.md
https://github.com/clearcontainers/runtime/blob/master/docs/ubuntu-installation-guide.md
https://github.com/clearcontainers/tests/blob/265e91f8fb0a5135eef215c8e780e92c39bfbe42/.ci/install_runtime.sh#L63

PR #773, opened 2017-10-28
Fix clear-containers service unit

frioux:
I'm not positive this is correct.  The old version (where there were two ExecStarts) errored immediately.  This version works, but causes `sudo systemctl docker start` to block forever, (though it does actually still work.)

### DIFF
--- a/docs/ubuntu-installation-guide.md
+++ b/docs/ubuntu-installation-guide.md
@@ -36,7 +36,7 @@ $ sudo -E apt-get -y install cc-runtime cc-proxy cc-shim
 $ sudo mkdir -p /etc/systemd/system/docker.service.d/
 $ cat <<EOF | sudo tee /etc/systemd/system/docker.service.d/clear-containers.conf
 [Service]
-ExecStart=
+Type=oneshot
 ExecStart=/usr/bin/dockerd -D --add-runtime cc-runtime=/usr/bin/cc-runtime --default-runtime=cc-runtime
 EOF
 ```


jodh-intel:
Hi @frioux - thanks for raising, but I don't think this is correct. `oneshot` won't work for docker as it is a long-running process. Also, the reason we have a seemingly empty `ExecStart=` is to reset the default docker service's `ExecStart=` which is required before we then specify a new value (the 2nd `ExecStart=`).

Can you provide further details of what errors you saw originally?

frioux:
Thanks for the quick response @jodh-intel.  What made me think this was the following:

```
Oct 28 09:42:41 zfp systemd[1]: docker.service: Service has more than one ExecStart= setting, which is only allowed for Type=oneshot services. Refusing.
```

I am attaching an (almost) full log of when I was playing with clear containers.  aufs implies clear, and overlay2 implies native cgroups and namespaces.
[big.txt](https://github.com/clearcontainers/runtime/files/1428332/big.txt).  I can try more if you'd like, but it's a pretty big hassle as I run docker all the time for local tools.  I'm on ubuntu 17.10.

jodh-intel:
Hi @frioux - it sounds like your systemd config is invalid. You should have:

- a `/lib/systemd/system/docker.service` with a single `ExecStart=` line.
- a `/etc/systemd/system/docker.service.d/clear-containers.conf` containing two `ExecStart=` lines, where the first one doesn't have anything after the equals sign (as specified in step (3) in https://github.com/clearcontainers/runtime/blob/master/docs/ubuntu-installation-guide.md).

jodh-intel:
Hi @frioux - Yes, you are right that the doc doesn't mention `/lib/systemd/system/docker.service` explicitly. But that will be created when you install the `docker-ce` package in step (1). Most users won't ever need to touch that file. In fact, our install docs assume (where possible) that all distro-packaged files as pristine/sacred/read-only, so that is why we use the systemd drop-in snippet approach by creating `/etc/systemd/system/docker.service.d/clear-containers.conf` to override the default (distro-packaged) docker behaviour.

For an alternative we are considering, take a look at #754 and #763.

Thanks - yes, please do confirm that with an unmodified  `/lib/systemd/system/docker.service` you can now create a Clear Container.

frioux:
So I just installed docker-ce afresh and here is the default ` /lib/systemd/system/docker.service`:

```
[Unit]
Description=Docker Application Container Engine
Documentation=https://docs.docker.com
After=network-online.target docker.socket firewalld.service
Wants=network-online.target
Requires=docker.socket

[Service]
Type=notify
# the default is not to use systemd for cgroups because the delegate issues still
# exists and systemd currently does not support the cgroup feature set required
# for containers run by docker
ExecStart=/usr/bin/dockerd -H fd://
ExecReload=/bin/kill -s HUP $MAINPID
LimitNOFILE=1048576
# Having non-zero Limit*s causes performance problems due to accounting overhead
# in the kernel. We recommend using cgroups to do container-local accounting.
LimitNPROC=infinity
LimitCORE=infinity
# Uncomment TasksMax if your systemd version supports it.
# Only systemd 226 and above support this version.
TasksMax=infinity
TimeoutStartSec=0
# set delegate yes so that systemd does not reset the cgroups of docker containers
Delegate=yes
# kill only the docker process, not all processes in the cgroup
KillMode=process
# restart the docker process if it exits prematurely
Restart=on-failure
StartLimitBurst=3
StartLimitInterval=60s

[Install]
WantedBy=multi-user.target
```

*But* it does seem to be working anyway.  Thanks; I'll close this now.

jodh-intel:
\o/. Thanks for getting back to use @frioux - glad it's all working now! :smile:

PR #775, opened 2017-10-30
Improve test info msgs

jodh-intel:


main_test.go
```diff
@@ -103,6 +103,12 @@ func init() {
 
 	fmt.Printf("INFO: test directory is %v\n", testDir)
 
+	fmt.Printf("INFO: ensuring docker is running\n")
+	_, err = runCommand([]string{"docker", "version"})
+	if err != nil {
+		panic(fmt.Sprintf("ERROR: docker daemon is not running: %v", err))
```
grahamwhaley: nit: technically, I guess it might not just be that the daemon is not running, it could be for instance that docker is not installed at all.  `Failed to get docker version: %v` or `Failed to launch docker command: %v` may be more accurate?

sameo:
LGTM

PR #778, opened 2017-10-30
Networking: Revendor netlink and virtcontainers to support macvtap

### LINKED ISSUE #708: Revendor vishwananda netlink to support index specification
# Description of problem

We need the ability to specify the index when creating the macvtap interfaces due to a kernel limitation
https://patchwork.ozlabs.org/patch/612683/

```
When creating macvtaps that are expected to have the same ifindex
in different network namespaces, only the first one will succeed.
The others will fail with a sysfs_warn_dup warning due to them trying
to create the following sysfs link (with 'NN' the ifindex of macvtapX):

/sys/class/macvtap/tapNN -> /sys/devices/virtual/net/macvtapX/tapNN

This is reproducible by running the following commands:

ip netns add ns1
ip netns add ns2
ip link add veth0 type veth peer name veth1
ip link set veth0 netns ns1
ip link set veth1 netns ns2
ip netns exec ns1 ip l add link veth0 macvtap0 type macvtap
ip netns exec ns2 ip l add link veth1 macvtap1 type macvtap

The last command will fail with "RTNETLINK answers: File exists" (along
with the kernel warning) but retrying it will work because the ifindex
was incremented.

The 'net' device class is isolated between network namespaces so each
one has its own hierarchy of net devices.
This isn't the case for the 'macvtap' device class.
The problem occurs half-way through the netdev registration, when
`macvtap_device_event` is called-back to create the 'tapNN' macvtap
class device under the 'macvtapX' net class device.

This patch adds namespace support the the 'macvtap' device class so
that /sys/class/macvtap is no longer shared between net namespaces.

However, doing this has the side effect of changing
/sys/devices/virtual/net/macvtapX/tapNN  into
/sys/devices/virtual/net/macvtapX/macvtap/tapNN

This is due to Commit 24b1442 ("Driver-core: Always create class
directories for classses that support namespaces.")
```

We need the following PR to land in netlink before we revendor
https://github.com/vishvananda/netlink/pull/283

### LINKED ISSUE #709: Add support for macvtap based network interconnection model
Needs revendoring of virtcontainers https://github.com/containers/virtcontainers/pull/393 and (qemu and netlink)

Networking: Add support for multi-queue mactap

Provide multiple methods to connect the Virtual machine
to the container network. The current implementation allows
this to be chosen at a node level. In the future we can
enhance this to be dynamic, where the container interface
type is used to determine the optimal interconnection method.

Add support for multi-queue macvtap as an alternate means to
connect the container network interface to the virtual machine.

mcastelino:
Revendor netlink and virtcontainers to switch the default network connection method to use macvtap.

Fixes #708
Fixes #709

### DIFF
--- a/Gopkg.lock
+++ b/Gopkg.lock
@@ -30,7 +30,7 @@
 [[projects]]
   name = "github.com/containers/virtcontainers"
   packages = [".","pkg/cni","pkg/hyperstart","pkg/oci","pkg/uuid","pkg/vcMock"]
-  revision = "e8b0c3a6c9fd7691e37a6aa73ed6af4c42f42ad1"
+  revision = "c3c42b5690ce384a26dcb8ac59e922a20dd0ac7a"
 
 [[projects]]
   branch = "master"
@@ -96,7 +96,7 @@
 [[projects]]
   name = "github.com/vishvananda/netlink"
   packages = [".","nl"]
-  revision = "177f1ceba557262b3f1c3aba4df93a29199fb4eb"
+  revision = "b7fbf1f5291ecf8ae5179d3202e914cb98cfe400"
 
 [[projects]]
   name = "github.com/vishvananda/netns"
@@ -116,6 +116,6 @@
 [solve-meta]
   analyzer-name = "dep"
   analyzer-version = 1
-  inputs-digest = "3c1e5894ba73555107971ec0b7f8e27019e5f1f8669f8e25ae6fd8f3b345c110"
+  inputs-digest = "d420f5b96ebe118ff1650e702cd485bebe83967b1053d7ae4a999966ba765ac9"
   solver-name = "gps-cdcl"
   solver-version = 1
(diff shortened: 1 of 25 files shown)


sboeuf:
@mcastelino I think we should stick the runtime vendoring to specific versions as we did for virtcontainers, WDYT ?

mcastelino:
> I think we should stick the runtime vendoring to specific versions as we did for virtcontainers, WDYT ?
I was thinking about that. The only downside (or upside) if you do that is that the revendoring becomes a more manual (but thoughtful process). I can do that in this PR if you want.

sboeuf:
@mcastelino, no I was mentioning this because I thought it would be good to have it. But I think this should be done in a separate PR so that it's not hidden behind netlink and virtcontainers revendoring.

sboeuf:
LGTM

devimc:
lgtm

PR #780, opened 2017-10-31
Vendoring: Switch to fixed revision vendoring

### LINKED ISSUE #779: Switch runtime to using specific revisions for vendored code
# Description of problem

virtcontainers has already switched to vendoring specific revisions of dependencies. 

Switch the runtime to also adopt the same. 

# Expected result

This allows an user to do perform 
```
rm -rf vendor 
dep ensure
```
without having to worry about unwanted changes being picked up.

The downside is that the revendoring becomes manual. It also requires that runtime dependency revisions match that the exact revision used by virtcontainers (which may be a good thing).

mcastelino:
virtcontainers has already switched to vendoring specific revisions of dependencies.
Switch the runtime to also adopt the same.

This allows an user to do perform

rm -rf vendor
dep ensure

without having to worry about unwanted changes being picked up.

The downside is that the revendoring becomes manual. It also requires that runtime dependency
revisions match that the exact revision used by virtcontainers (which may be a good thing).

Fixes #779

### DIFF
--- a/Gopkg.lock
+++ b/Gopkg.lock
@@ -7,7 +7,6 @@
   revision = "b2a8a2ecea1318ae28a0861191ec385532491b5a"
 
 [[projects]]
-  branch = "master"
   name = "github.com/BurntSushi/toml"
   packages = ["."]
   revision = "a368813c5e648fee92e5f6c30e3944ff9d5e8895"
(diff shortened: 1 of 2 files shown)


sameo:
LGTM

jodh-intel:
lgtm

PR #830, opened 2017-11-24
vendor: update virtcontainers

devimc:
this new version of virtcontainers has support for hotplugging
in Qemu Q35

shortlog:
aa2ba14 Remove redundant "if" check
37dc3c6 cni: Assign the netpair for virtual endpoints after scan
8f846a2 cni: Fix bug while updating endpoints after scanning net namespace
eb39def qemu: Provide unique QMP sockets names
6026603 qemu: Don't wait for the end of the VM
eeacf8b qemu: Fix disconnectCh issue and simplify it
711783f agent: hyperstart: Tear down properly the pod
b20ba87 pod: Improve pod create performances
d449ece api: Refactor RunPod()
c3663b2 pod: Run post-start hooks outside netns
42df5a6 hypervisor: add hot plugging support for Qemu Q35

fixes #829

Signed-off-by: Julio Montes <julio.montes@intel.com>

PR #832, opened 2017-11-24
hotplug: add default_bridges to configuration file

### LINKED ISSUE #831: add option in the configuration file to specify the number PCI bridges
allow user specify how many PCI bridges include the VM, PCI bridges are used to hot plug devices

devimc:
default_bridges allows user specify how many PCI bridges
must be included in the VM, this allow to hot plug until
~150 devices

fixes #831

### DIFF
--- a/config.go
+++ b/config.go
@@ -57,6 +57,9 @@ const (
 
 	// supported agent component types
 	hyperstartAgentTableType = "hyperstart"
+
+	// the maximum amount of PCI bridges that can be cold plugged in a VM
+	maxPCIBridges uint32 = 5
 )
 
 var (
@@ -82,6 +85,7 @@ type hypervisor struct {
 	MachineType           string `toml:"machine_type"`
 	DefaultVCPUs          int32  `toml:"default_vcpus"`
 	DefaultMemSz          uint32 `toml:"default_memory"`
+	DefaultBridges        uint32 `toml:"default_bridges"`
 	DisableBlockDeviceUse bool   `toml:"disable_block_device_use"`
 	MemPrealloc           bool   `toml:"enable_mem_prealloc"`
 	HugePages             bool   `toml:"enable_hugepages"`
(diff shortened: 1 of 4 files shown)


devimc:
depends on https://github.com/clearcontainers/runtime/pull/830

config.go
```diff
@@ -172,6 +173,20 @@ func (h hypervisor) defaultMemSz() uint32 {
 	return h.DefaultMemSz
 }
 
+func (h hypervisor) defaultBridges() uint32 {
+	if h.DefaultBridges <= 0 {
```
jodh-intel: - This should be `if h.DefaultBridges == 0` since the var is an unsigned.
- Can you add a test for this function?

config.go
```diff
@@ -172,6 +173,20 @@ func (h hypervisor) defaultMemSz() uint32 {
...
+		return defaultBridgesCount
+	}
+
+	// To allow other devices be plugged directly in the pci(e).0
+	// until 10 bridges can be plugged per POD/VM.
+	if h.DefaultBridges > 10 {
```
jodh-intel: - I don't understand this limit. What if I want `11` bridges for my pod?
- Can you either expand the comment to explain this in more detail, or mention that more details are provided in the config file?
- How about creating a const for the `10`?
- If we return `10` here, I think we should emit a log message explaining what is happening since we're overriding what the user requested in the config file.

devimc: Actually , I can plug until 14 bridges in Q35 and 8 in PC, It could be an limitation not documented on Qemu or We have an issue in the kernel.
Anyway, I'll change this to 5 and document this limitation

config.go
devimc: yes, Thanks @jodh-intel

jodh-intel:
lgtm.

There is a test failure (which looks like somehow `defaultBridgesCount` is getting set to zero?)

PR #857, opened 2017-12-13
Networking: Add support for multiqueue tap

mcastelino:
Add support for multiqueue tap and unify the handling of both
macvtap and tap based connection methods.

This is achieved by the revendoring of the virtcontainers
and qemu packages. There is no code change in the runtime itself.


sboeuf:
LGTM

sameo:
LGTM

PR #863, opened 2017-12-14
Add support to BC-Linux system

tom0392:
Add a new script of bclinux-setup.sh to add support to the [BC-Linux](https://www.bclinux.org/) system.

Signed-off-by: tangzhongrui tangzhongrui@cmss.chinamobile.com

### DIFF
--- a/docs/bclinux-installation-guide.md
+++ b/docs/bclinux-installation-guide.md
@@ -1,0 +1,8 @@
+# Installing Intel® Clear Containers 3.0 on BC-Linux* version 7
+## Introduction of BC-Linux
+[BC-Linux](https://www.bclinux.org/) is an enterprise Linux operating system developed by China Mobile for cloud computing and big data users.   
+BC-Linux has developed some common server operating systems based on the CentOS OS Open Source Community Edition. For example, BC-Linux 7.3 is based on the same source code as Red Hat Enterprise RHEL 7.3 and maintains 100% compatibility with RHEL.   
+Additionally, BCLinux has released some customized versions of the server OS for different scenarios. These customized versions simplify the common BC-Linux 7 system and integrate the packages required for the cloud computing and big data scenarios. BCLinux-for-DCOS is a good representative. It uses a customized high-version kernel, fixes the security vulnerabilities in the low-version kernel, and adds multiple optimization patches for the container.
+
+## Installation steps
+Clear Containers 3.0 is available for BC-Linux 7.3, 7.2, and BCLinux-for-DCOS-V1.0. BC-Linux is basically a variant of CentOS so users can refer to [docs/centos-installiation-guide.md](https://github.com/clearcontainers/runtime/blob/master/docs/centos-installation-guide.md) to install Clear Containers 3.0 on BC-Linux* version 7.
--- a/installation/centos-setup.sh
+++ b/installation/centos-setup.sh
@@ -19,8 +19,22 @@
 #  Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA.
 
 # Description: This script installs Clear Containers on a
-#   CentOS 7 system.
+#   CentOS 7 and other variants of CentOS system.
 #
+
+# add the -h option to list systems which the script supports
+while getopts "h" opt
+do
+	case $opt in
+	"h")
+		echo -e "The script is designed to work on the following systems:\n(1) centos\n(2) bclinux"
+		exit
+		;;
+	*)
+                echo "Wrong Options"
+		;;	
+	esac
+done	
 
 # all errors are fatal
 set -e
@@ -68,12 +82,12 @@ source /etc/os-release
 
 major_version=$(echo "${VERSION_ID}"|cut -d\. -f1)
 
-if [ "${os_distribution}" = centos ]
+if [ "${os_distribution}" = centos -o "${os_distribution}" = bclinux ]
 then
     distro="CentOS"
 else
     echo >&2 "ERROR: Unrecognised distribution: ${os_distribution}"
-    echo >&2 "ERROR: This script is designed to work on CentOS systems only."
+    echo >&2 "ERROR: This script is designed to work on CentOS system and also applies to other variants of CentOS system."
     exit 1
 fi
 
--- a/docs/centos-installation-guide.md
+++ b/docs/centos-installation-guide.md
@@ -61,6 +61,11 @@ Note:
   installation to the file `typescript`. This is useful for administrators
   to see the changes made and for use when debugging issues.
 
+- The installation/centos-setup.sh script works for CentOS and selects other variants. To obtain a list of supported variants, please run:
+```
+$ installation/centos-setup.sh -h
+```
+
 ## The following steps show how to verify the Clear Containers 3.X installation is successful:
 
 1. Check the `cc-runtime` version:


jodh-intel:
Hi - firstly, thanks for raising this PR.

The new script is almost identical to the CentOS installation script so I'm assuming BC-Linux is a variant of CentOS Linux. If I'm right, I think a better approach would be to make the following changes:

- Update https://github.com/clearcontainers/runtime/blob/master/installation/centos-setup.sh to accept an optional argument which is a CentOS variant. If that variant is recognised by the script, the installation is allowed to proceed.
- The `centos-setup.sh` script can then define variant-specific code blocks (such as `[ "${os_distribution}" = bclinux ] && do_something_specific_to_bc_linux`).
- Update https://github.com/clearcontainers/runtime/blob/master/docs/centos-installation-guide.md to explain that the script supports a limited number of variants and that to get a list of supported distro variants, the user should run:
  ```
  $ ./installation/centos-setup.sh -h
  ```
  This should show some text explaining what the script is for along with the list of supported variants (which would be `centos` and `bclinux`).
   ```

installation/bclinux-setup.sh
```diff
@@ -0,0 +1,146 @@
...
+
+pkgs+=" gcc-c++"
+
+source /etc/os-release
+
+major_version=$(echo "${VERSION_ID}"|cut -d\. -f1)
+
+if [ "${os_distribution}" = bclinux -o "${os_distribution}" = centos ]
+then
+    distro_exapnd="bclinux"
```
jodh-intel: What is this line for? It doesn't appear to be used? Also, should that be `distro_expand=` if it is used somehow?

installation/bclinux-setup.sh
```diff
@@ -0,0 +1,146 @@
...
+sudo yum -y update
+eval sudo yum -y install "${pkgs}"
+
+sudo yum groupinstall -y 'Development Tools'
+
+pushd "${deps_dir}"
+
+# Install pre-requisites for gcc
+gmp_file="gmp-${gmp_version}.tar.bz2"
+#curl -L -O "https://gcc.gnu.org/pub/gcc/infrastructure/${gmp_file}"
```
jodh-intel: All the `curl` lines are commented out. I don't understand why that is. I'm assuming it's either a mistake or that these lines should be deleted entirely?

grahamwhaley:
The F26 CI failure is unrelated (as I'd expect) - looks like a network timeout - so we don't need to kick that just now as I suspect this PR will get some rework and automatically trigger a rebuild anyhow.
Just for reference:
```
 policycoreutils-python3      x86_64 2.6-6.fc26                   updates 1.8 M
12:47:15  popt     Error: Error downloading packages:
12:47:15   Curl error (28): Timeout was reached for https://mirrors.fedoraproject.org/metalink?repo=updates-released-f26&arch=x86_64 [Connection timed out after 30002 milliseconds]
12:47:15                     x86_64 1.16-12.fc26                 updates  58 k
```

installation/bclinux-setup.sh
```diff
@@ -0,0 +1,146 @@
+#!/bin/bash
```
devimc: would be nice to have a bclinux-installation-guide.md in https://github.com/clearcontainers/runtime/tree/master/docs
to include it in the wiki

installation/bclinux-setup.sh
```diff
@@ -0,0 +1,146 @@
...
+sudo yum-config-manager --add-repo https://download.docker.com/linux/centos/docker-ce.repo
+sudo yum -y install docker-ce
+
+# Install Clear Containers components and their dependencies
+sudo yum-config-manager --add-repo "${cc_repo_url}"
+sudo yum -y install cc-runtime cc-proxy cc-shim linux-container clear-containers-image
+
+# Override runtime configuration to use hypervisor from prefix_dir
+# rather than the OBS default values.
+sudo -E prefix_dir="${prefix_dir}" sed -i -e \
```
devimc: this is interesting,  where is defined ```prefix_dir```? it this something specific of bclinux?

devimc: ohh I see, it's defined in ```installation-setup.sh```

installation/bclinux-setup.sh
tom0392: Hi,@jodh-intel, you are right. It doesn't be used now. I added this line because I thought we could use it later.

installation/bclinux-setup.sh
tom0392: I‘m sorry that it's a mistake. I will update that script to correct it. Everytime I run the script, some large size packages can't be downloaded completely which will cause installation errors. So I  download them manually then copy them to the directory under ./runtime/.   At last, re-run the script after commenting out these lines. In this way the script can be installed correctly.

installation/bclinux-setup.sh
tom0392: Hi,@devimc,I have written a document of bclinux-installation which I 'm going to submit.

tom0392:
According to suggestions from @jodh-intel and @devimc , I have corrected the errors in the script and have written a document of bclinux-installation-guide in this submission. Thanks.

grahamwhaley:
Hi @tom0392 Looking good :-)
Did you consider merging your `bclinux-setup.sh` changes into the existing https://github.com/clearcontainers/runtime/blob/master/installation/centos-setup.sh like @jodh-intel suggested?
I think for the script this is probably best?
Also for the documentation - that looks to be 90+% a copy - so maybe the bc-linux document can just have the BC specific introduction, and then reference the centos doc for more details?

tom0392:
Hi,@grahamwhaley and @jodh-intel , I think it over seriously. That's great and more reasonable. I will try to update https://github.com/clearcontainers/runtime/blob/master/installation/centos-setup.sh to add support to BC-Linux system and update https://github.com/clearcontainers/runtime/blob/master/docs/centos-installation-guide.md accordingly. Thanks for all your comments.

jodh-intel:
Hi @tom0392 - Thanks for updating the PR.

However, I still feel that there is duplication here since `docs/bclinux-installation-guide.md` is almost identical in content to the `docs/centos-installation-guide.md` document. See again my comments in https://github.com/clearcontainers/runtime/pull/863#issuecomment-351703812.

My suggestion would be to:

- Update the `docs/centos-installation-guide.md` document to state that:
  > The `installation/centos-setup.sh` script works for CentOS and selected other variants. To obtain a list of supported variants, please run:
  > ```
  > $ installation/centos-setup.sh -h
  > ```
- Add the `-h` option to `installation/centos-setup.sh` which should list that it supports:
  - CentOS
  - BC-Linux
- If you want, you could also create a very small "stub" document called `docs/bclinux-installation-guide.md` that simply:
  - States that BC-Linux is a variant of CentOS.
  - Refers the user to `docs/centos-installation-guide.md`.
  In other words, that document would include no technical content - that would all be handled in the CentOS installation guide document.

@klynnrif, @rcaballeromx - does the above sound good to you?

rcaballeromx:
@jodh-intel I agree with your suggestion. Particularly since the variants are listed via the -h option instead of in the document. That way, future supported variants can be added with minimal change to the docs.

tom0392:
@jodh-intel Thanks! I have pushed changes for your comments. Can you take a look at this?

jodh-intel:
Thanks @tom0392 - Just a couple of minor comments. I should also ask before this lands... Are you happy to maintain `docs/bclinux-installation-guide.md` and the BC-Linux parts of `installation/centos-setup.sh`? I suspect there won't be a lot of work to do, however, if a new version of BC-Linux (or CentOS) is released, we'll need to worth together to ensure that `installation/centos-setup.sh` is still able to handle the setup for both BC-Linux and CentOS.

/cc @grahamwhaley

@rcaballeromx / @klynnrif - could you take a look at the document changes here please?

docs/bclinux-installation-guide.md
```diff
@@ -0,0 +1,8 @@
+# Installing Intel® Clear Containers 3.0 on BCLinux* version 7
+## Introduction of BC-Linux
+[BC-Linux](https://www.bclinux.org/)（BigCloud Enterprise Linux） is an enterprise Linux operating system developed by China Mobile for cloud computing and big data users.
```
klynnrif: Remove extra space: ...(BigCloud Enterprise Linux) is an...

docs/bclinux-installation-guide.md
```diff
@@ -0,0 +1,8 @@
...
+BC-Linux has developed some common server operating systems based on the CentOS OS Open Source Community Edition. For example, BC-Linux 7.3 is based on the same source code as Red Hat Enterprise RHEL 7.3 and maintains 100% compatibility with RHEL.
+In addition, BCLinux has released some customized versions of server OS for different scenarios. BCLinux-for-DCOS is a representative product of them. It greatly simplifies the common BC-Linux 7 system and integrates the packages required for the container and PaaS Platform. It uses a customized high-version kernel, fixes the security vulnerabilities in the low-version kernel, and adds multiple optimization patches for the container.
```
klynnrif: Additionally, BCLinux has released some customized versions of the server OS for different scenarios (e.g. BCLinux-for-DCOS). These customized versions simplify the common BC-Linux 7 system and integrate the packages required for the container and PaaS Platform. Each version uses a customized high-version kernel, fixes the security vulnerabilities in the low-version kernel, and adds multiple optimization patches for the container.

docs/bclinux-installation-guide.md
```diff
@@ -0,0 +1,8 @@
...
+
+## Installation steps
+Clear Containers 3.0 is available for BC-Linux 7.3, 7.2 and BCLinux-for-DCOS-V1.0. For now, BC-Linux is a variant of CentOS in a sense, so users can refer to docs/centos-installation-guide.md to install Clear Containers 3.0 on BCLinux* version 7.
```
klynnrif: Clear Containers 3.0 is available for BC-Linux 7.3, 7.2, and BCLinux-for-DCOS-V1.0. BC-Linux is basically a variant of CentOS so users can refer to docs/centos-installiation-guide.md to install Clear Containers 3.0 on BCLinux* version 7.

tom0392:
@jodh-intel Thanks. It's my great honor and I am very happy to do this job. I will try my best to maintain docs/bclinux-installation-guide.md and the BC-Linux parts of installation/centos-setup.sh. Happy New Year :)

docs/bclinux-installation-guide.md
tom0392: Actually, we have different customized versions and BCLinux-for-DCOS is one of them. Only BCLinux-for-DCOS integrates the packages required for the container and PaaS Platform, and it uses a customized high-version kernel.
So, @klynnrif，how about the following changes:
Additionally, BCLinux has released some customized versions of the server OS for different scenarios. These customized versions simplify the common BC-Linux 7 system and integrate the packages required for the cloud computing and big data scenarios. BCLinux-for-DCOS is a good representative.  It uses a customized high-version kernel, fixes the security vulnerabilities in the low-version kernel, and adds multiple optimization patches for the container.

docs/bclinux-installation-guide.md
tom0392: OK, There is duplication here and I will remove it. Thanks.

docs/bclinux-installation-guide.md
klynnrif: @tom0392 lgtm

docs/bclinux-installation-guide.md
```diff
@@ -0,0 +1,8 @@
...
+## Introduction of BC-Linux
+[BC-Linux](https://www.bclinux.org/) is an enterprise Linux operating system developed by China Mobile for cloud computing and big data users.
+BC-Linux has developed some common server operating systems based on the CentOS OS Open Source Community Edition. For example, BC-Linux 7.3 is based on the same source code as Red Hat Enterprise RHEL 7.3 and maintains 100% compatibility with RHEL.
+Additionally, BCLinux has released some customized versions of the server OS for different scenarios. These customized versions simplify the common BC-Linux 7 system and integrate the packages required for the cloud computing and big data scenarios. BCLinux-for-DCOS is a good representative. It uses a customized high-version kernel, fixes the security vulnerabilities in the low-version kernel, and adds multiple optimization patches for the container.
+
+## Installation steps
+Clear Containers 3.0 is available for BC-Linux 7.3, 7.2, and BCLinux-for-DCOS-V1.0. BC-Linux is basically a variant of CentOS so users can refer to docs/centos-installiation-guide.md to install Clear Containers 3.0 on BC-Linux* version 7.
```
jodh-intel: Ah - I've just noticed that "docs/centos-installiation-guide.md" needs to be a link.

jodh-intel:
Hi @tom0392 - once you've added the link to the doc, this can be merged (the Fedora 26 issue is affecting all of the open PRs and since this PR is only adding documentation and a script that isn't being run by the CI, we can ignore the F26 error here).

installation/centos-setup.sh
```diff
@@ -68,12 +82,12 @@ source /etc/os-release
...
 major_version=$(echo "${VERSION_ID}"|cut -d\. -f1)
 
-if [ "${os_distribution}" = centos ]
+if [ "${os_distribution}" = centos -o "${os_distribution}" = bclinux ]
 then
     distro="CentOS"
 else
     echo >&2 "ERROR: Unrecognised distribution: ${os_distribution}"
-    echo >&2 "ERROR: This script is designed to work on CentOS systems only."
+    echo >&2 "ERROR: This script is designed to work on CentOS system and also applies to other variants of Centos system."
```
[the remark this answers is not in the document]
tom0392: This script is designed to work on CentOS system and also applies to other variants of CentOS system. 
Am I right?

installation/centos-setup.sh
```diff
@@ -19,9 +19,23 @@
...
-#   CentOS 7 system.
+#   CentOS 7 and other variants of CentOS system.
 #
 
+# add the -h option to list systems which the script supports
+while getopts "h" opt
+do
+	case $opt in
+	"h")
+		echo -e "The script is designed to work on the following systems:\n(1) CenOS\n(2) BC-Linux"
```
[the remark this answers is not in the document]
tom0392: Ok，that's great.

jodh-intel:
Thanks @tom0392 - merging...

PR #873, opened 2017-12-20
exec: Fix "docker exec --user"

### LINKED ISSUE #867: exec should not use the OCI `Username` value
`exec.go` uses `params.ociProcess.User.Username`. This is at best confusing since `Username` is a [Windows-specific](https://github.com/opencontainers/runtime-spec/blob/master/config.md#windows-user) value.

Once the changes on https://github.com/containers/virtcontainers/pull/522 land in the virtcontainers `master` branch, we need to re-vendor that into the runtime then change `execute()` and `generateExecParams()` to only use `params.ociProcess.User.uid`.

See:

- https://github.com/containers/virtcontainers/pull/522#discussion_r157538887

jodh-intel:
Fix running `docker exec` with the `--user=` option. Previously,
specifying `--user=` with any user (or user and group) would be ignored
and the workload would run as `root` (`UID`+`GID`). This was because the
runtime was setting the user to run the workload from the `Username`
field in the OCI `Process.User` struct. However, that field is
Windows-only so the value was always zero, denoting `root`.

Fixes #867.

### DIFF
--- a/exec.go
+++ b/exec.go
@@ -225,11 +225,17 @@ func execute(context *cli.Context) error {
 		return err
 	}
 
+	user := fmt.Sprintf("%d:%d", params.ociProcess.User.UID, params.ociProcess.User.GID)
+
+	if params.ociProcess.User.Username != "" {
+		user = params.ociProcess.User.Username
+	}
+
 	cmd := vc.Cmd{
 		Args:        params.ociProcess.Args,
 		Envs:        envVars,
 		WorkDir:     params.ociProcess.Cwd,
-		User:        params.ociProcess.User.Username,
+		User:        user,
 		Interactive: params.ociProcess.Terminal,
 		Console:     consolePath,
 		Detach:      noNeedForOutput(params.detach, params.ociProcess.Terminal),
--- a/exec.go
+++ b/exec.go
@@ -157,6 +157,11 @@ func generateExecParams(context *cli.Context, specProcess *oci.CompatOCIProcess)
 		// Override user
 		if context.String("user") != "" {
 			params.ociProcess.User = specs.User{
+				// This field is a Windows-only field
+				// according to the specification. However, it
+				// is abused here to allow the username
+				// specified in the OCI runtime configuration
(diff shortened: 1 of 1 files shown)


exec.go
```diff
@@ -225,11 +225,17 @@ func execute(context *cli.Context) error {
 		return err
 	}
 
+	user := fmt.Sprintf("%d:%d", params.ociProcess.User.UID, params.ociProcess.User.GID)
+
+	if params.ociProcess.User.Username != "" {
+		user = params.ociProcess.User.Username
```
sboeuf: Are you sure we want to default to this in case we don't have anything provided. I think we should force `0:0`. This should be cleaner than to expect this `Username` to be 0 in order to set the user as root.

jodh-intel: I'm not sure I understand - the code will default the the UID+GID provided by the OCI config file, but if the user specifies `--user=`, we'll honour that instead. But note that the `--user=` case only applies for running `cc-runtime` "standalone" and has no impact when run under docker.

sboeuf: Oh yes I get it. I have misread the code...

sboeuf:
LGTM

sboeuf:
@jodh-intel CI issue:
```
/home/jenkins/jenkins_slave/workspace/clear-containers-runtime-fedora-26-PR/go/src/github.com/clearcontainers/runtime/exec.go:164:18: "overriden" is a misspelling of "overridden"
```

jodh-intel:
Ta - branch updated.

jodh-intel:
Note that we need to add a test for this at some point (soon!! :), so I've raised https://github.com/clearcontainers/tests/issues/805.

@jcvenegas - it would be good to get this PR in the next release btw.

jodh-intel:
https://github.com/clearcontainers/tests/pull/808 raised for a new test for this fix.

jodh-intel:
Hi @grahamwhaley - could you take a look at this?

jodh-intel:
ta.

Issue #795, opened 2017-11-07
Use one proxy per VM


jodh-intel:
Modify runtime to launch one proxy per VM. This will improve our HA story since there will no longer be a single point of failure on the system (the system-level instance of `cc-proxy`).

# Plan (ordered steps)

- [x] Update the proxy to allow 1 proxy / pod, alongside the system-level systemd `cc-proxy` service (https://github.com/clearcontainers/proxy/pull/167).
  At this point:
  - multiple instances of `cc-proxy` will be *able* to co-exist with the system-level `cc-proxy` systemd service.
- [x] Update virtcontainers to spawn 1 proxy / pod (https://github.com/containers/virtcontainers/pull/483).
- [x] Re-vendor the virtcontainers changes into the runtime (#835).
  At this point:
  - the system-level `cc-proxy` systemd instance won't be being used (although it can continue to run without interfering with the pod-specific `cc-proxy` instances).
  - we will lose the KSM feature of `cc-proxy` because since all the proxy requests will be handled by the pod-specific `cc-proxy` instances and since we don't want multiple proxies fighting over KSM kernel settings, the KSM code will never be run.
- [ ] Create a daemon to replace the KSM functionality in `cc-proxy` (https://github.com/clearcontainers/proxy/issues/168).
- [ ] Remove the following features from the proxy:
  - KSM.
  - socket activation.
  - handling code for >1 pods.
- [ ] Update packaging for the proxy to do the following:
  ```
   $ sudo systemctl stop cc-proxy.socket
   $ sudo systemctl stop cc-proxy.service
   $ sudo rm /lib/systemd/system/cc-proxy.socket
   $ sudo rm /lib/systemd/system/cc-proxy.service
   $ sudo systemctl daemon-reload
  ```
- [ ] Release updated packages for proxy, runtime and the new KSM daemon.
  At this point, there will no longer be a system-level `cc-proxy` instance.

dvoytik:
Hi @jodh-intel,

Overall IMHO this make sense.

Btw, I guess https://github.com/clearcontainers/proxy/pull/107 still make sense with the above plan. Although the PR should be refactored a bit.

sboeuf:
@dvoytik yes this would still make sense, but with a lower priority. Having one proxy per VM would solve most of our issues. In case of a proxy crash we would only loose one pod.

PR #879, opened 2017-12-22
scripts: Improvements to cc-collect-data script

### LINKED ISSUE #878: Make improvements to cc-collect-data.sh
`cc-collect-data.sh` should log the timezone where it was run along with `docker version` details (since `docker info` doesn't show the client versions).

jodh-intel:
Add the timezone the script was run in along with `docker versions`
details (since `docker info` doesn't show client information).

Fixes #878.

### DIFF
--- a/data/cc-collect-data.sh.in
+++ b/data/cc-collect-data.sh.in
@@ -285,12 +285,21 @@ show_container_mgr_details()
 
 	if have_cmd "docker"; then
 		subheading "Docker"
+		run_cmd_and_show_quoted_output "docker version"
 		run_cmd_and_show_quoted_output "docker info"
+		run_cmd_and_show_quoted_output "systemctl show docker"
 	fi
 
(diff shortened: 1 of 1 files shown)


jodh-intel:
@sameo, @grahamwhaley - if you can think of anything else we can add (more k8s details possibly?), feel free to let me know ;)

grahamwhaley:
Only other thing that springs to mind, just from something I was doing earlier today, might be `systemctl [status|show] docker` ? Just a thought. But then we could go crazy and do that for all the components - and then we'd be overloaded with noise maybe.

jodh-intel:
@grahamwhaley - I think that's a great idea - service details added.

Note that I haven't added service details for the proxy as it's basically dormant and soon to be removed (see #795 and https://github.com/clearcontainers/packaging/pull/200).

devimc:
lgtm

PR #898, opened 2018-01-06
docs: Fix option name in limitations doc

### LINKED ISSUE #899: doc: Fix network option name in limitations doc
The limitations document incorrectly specified `--host=net`: it should be `--net=host`.

guangxuli:
docs: Fix option name in limitations doc.

Correct Seems that --host=net should be instead by --net=host.

Fixes #899.

### DIFF
--- a/docs/limitations.md
+++ b/docs/limitations.md
@@ -277,7 +277,7 @@ enabling use of `--net=host` when necessary.
 It should be noted, currently passing the `--net=host` option into a
 Clear Container may result in the Clear Container networking setup
 modifying, re-configuring and therefore possibly breaking the host
-networking setup. Do not use `--host=net` with Clear Containers.
+networking setup. Do not use `--net=host` with Clear Containers.
 
 #### `docker run --link`
 


docs/limitations.md
```diff
@@ -277,7 +277,7 @@ enabling use of `--net=host` when necessary.
 It should be noted, currently passing the `--net=host` option into a
 Clear Container may result in the Clear Container networking setup
 modifying, re-configuring and therefore possibly breaking the host
-networking setup. Do not use `--host=net` with Clear Containers.
+networking setup. Do not use `--net=host` with Clear Containers.
```
jodh-intel: Good catch @guangxuli! :smile:

In fact, I wonder if we should really be specifying the full option name too (`--network=`). What do you think @mcastelino  and @klynnrif ?

jodh-intel: Actually - let's just land this fix as it's consistent with what is already written. If we decide to only use the full option name, that can be dealt with on a separate PR.

jodh-intel:
@guangxuli - Before we land this, please could you change the commit itself slightly to include a subsystem and a brief description (see https://github.com/clearcontainers/runtime/blob/master/CONTRIBUTING.md#patch-format).

For this PR, I'd expect the commit to look something like this:

```
docs: Fix option name in limitations doc

Correct Seems that --host=net should be instead by --net=host.

Fixes #899.
```

Thanks again!

grahamwhaley:
agree with @jodh-intel - good catch @guangxuli (and apologies, that was probably my original typo :-) ) - if we get the commit rewritten to conform with the guildelines for the repo, we can merge.

guangxuli:
@jodh-intel Done. I sholuld read the guideline carefully.
@grahamwhaley  that's ok. :)

jodh-intel:
Hi @guangxuli - I can see the changes at the top of the PR, but the commit still shows the original value? Please could you update the commit itself and then force-push?

guangxuli:
@jodh-intel oh, yes, i haven't aware that. updated, thanks your patience.

jodh-intel:
Thanks @guangxuli!

lgtm

@klynnrif - please could you review?

PR #907, opened 2018-01-08
list: Don't fail if rootfs doesn't exist

### LINKED ISSUE #906: list should not fail on user lookup
Currently, if `list` is unable to display the UID of the rootfs, it will error. However, we should probably be more tolerant and display a `?` to denote the inability to display the user rather than failing the entire `list` command.

See #902.

jodh-intel:
If it is not possible to determine the owner of the rootfs, do not fail
the entire `list` operation, just mark the container owner field with a
question mark.

Fixes #906.

### DIFF
--- a/list.go
+++ b/list.go
@@ -323,12 +323,14 @@ func getContainers(context *cli.Context) ([]fullContainerState, error) {
 			ociState := oci.StatusToOCIState(container)
 			staleAssets := getStaleAssets(currentHypervisorDetails, latestHypervisorDetails)
 
-			uid, err := getDirOwner(container.RootFs)
-			if err != nil {
-				return nil, err
-			}
+			var owner string
 
-			owner := fmt.Sprintf("#%v", uid)
+			uid, err := getDirOwner(container.RootFs)
+			if err == nil {
+				owner = fmt.Sprintf("#%v", uid)
+			} else {
+				owner = "?"
+			}
 
 			s = append(s, fullContainerState{
 				containerState: containerState{
--- a/list_test.go
+++ b/list_test.go
@@ -430,6 +430,70 @@ func TestListGetContainers(t *testing.T) {
 	assert.Equal(state, []fullContainerState(nil))
 }
 
+func TestListGetContainersMissingRootfs(t *testing.T) {
+	assert := assert.New(t)
+
+	tmpdir, err := ioutil.TempDir(testDir, "")
+	assert.NoError(err)
+	defer os.RemoveAll(tmpdir)
+
+	pod := &vcMock.Pod{
+		MockID: testPodID,
+	}
+
+	rootfs := filepath.Join(tmpdir, "rootfs")
+
+	container := vc.ContainerStatus{
+		ID: pod.ID(),
+		State: vc.State{
+			State: vc.StateRunning,
+		},
+		PID: 1,
+
+		// rootfs specified, but it doesn't exist
+		RootFs: rootfs,
+	}
+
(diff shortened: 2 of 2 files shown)


chavafg:
lgtm

list_test.go
```diff
@@ -430,6 +430,70 @@ func TestListGetContainers(t *testing.T) {
...
+				ID: pod.ID(),
+				ContainersStatus: []vc.ContainerStatus{
+					container,
+				},
+			},
+		}, nil
+	}
+
+	defer func() {
+		testingImpl.ListPodFunc = nil
```
jcvenegas: dump question why this is need?

list.go
```diff
@@ -323,13 +323,15 @@ func getContainers(context *cli.Context) ([]fullContainerState, error) {
...
 
+			var owner string
+
 			uid, err := getDirOwner(container.RootFs)
-			if err != nil {
-				return nil, err
+			if err == nil {
+				owner = fmt.Sprintf("#%v", uid)
+			} else {
+				owner = "?"
```
jcvenegas: This is an error that started to happen recently,  I think we should be able always know the owner , so being this a not usual case can you log a warning and probably and  log the error to probably know if the file did not exist ?

jodh-intel: Yeah, I'm having second thoughts about whether we should do this now we've confirmed the CI failure (https://github.com/clearcontainers/jenkins/issues/32) is Fedora only.

What do you think @sameo, @grahamwhaley ? FTR, `runc` is "semi-lenient": it does not validate whether the container rootfs exists as it doesn't really need to care about that:

- https://github.com/opencontainers/runc/blob/master/list.go#L126..L139

Nominally, we do care because we use the container rootfs ownership to show who owns the container. However, we should really be employing a strategy similar to `runc` which sets the container ownership to the owner of the container directory below `runc`'s state directory. But `cc-runtime` cannot do that as it doesn't know where the virtcontainers root directory is (implementation detail).

As such, this is all slightly academic as `cc-runtime list` will always show the owner UID `0` currently.

The advantage of not failing on rootfs lookup is that `list` will show atleast partial information. Maybe we should hold off deciding until we understand the F26 failure further.

grahamwhaley: sure, mark it as WIP until we get some further info then OK?

list_test.go
jodh-intel: This resets the virtcontainers mock handler for `ListPod()`. See https://github.com/containers/virtcontainers/blob/master/pkg/vcMock/mock.go#L91..L99.

By resetting the handler to `nil`, we guarantee that any other test that invokes any virtcontainers API calls which either directly or indirectly call `ListPod()` will fail with an error from the virtcontainers mock package. The reason for all this being that some virtcontainers types are private, but the runtime's unit tests need to create particular test scenarios using those private data types. They also need to simulate previous calls to particular sequences of virtcontainers calls. Since the runtime uses the virtcontainers API heavily, we have a problem as we cannot contrive such scenarios using the main virtcontainers API calls (the runtime tests might not even have sufficient privs to run some of the virtcontainers API calls). As such, virtcontainers provides *two* implementations of its API - the "proper" / official one and a mock one that can be used for testing the caller of the API itself, as here.

list.go
jodh-intel: wfm...

jcvenegas:
Any update on this, I think is safe close it, or merge it but add a warning to our log.

jodh-intel:
Let's close it for now as it no longer seems to be required.

PR #926, opened 2018-01-15
config: honour VM config resources

### LINKED ISSUE #925: pod resources are not honoured in pod config file
# Description of problem

using constraints, pod vCPus and Memory are always 0 in kubernetes

run a container with kubernetes and take a look to the pod config file

```sh
kubectl --namespace=sandbox create -f ~/debian.yaml
cat /var/lib/virtcontainers/pods/$ID/config.json
... "VMConfig":{"VCPUs":0,"Memory":0}
```
_debian.yaml_

```yaml
apiVersion: v1
kind: Pod
metadata:
  name: test
spec:
  containers:
  - name: test1
    image: debian
    resources:
      limits:
        cpu: "4"
    args: [ nproc ]
    tty: true
  - name: test2
    image: debian
    args: [ echo, "bye world!" ]
  - name: test3
    image: debian
    resources:
      limits:
        cpu: "3"
    args:
    - bash
    stdin: true
    tty: true
```

do the same, but using docker

```sh
docker run -dti --cpus=5 debian bash
cat /var/lib/virtcontainers/pods/$ID/config.json
... "VMConfig":{"VCPUs":5,"Memory":0}
```

# Expected result

resources should not be 0 in the pod configuration

devimc:
VM config resources must be honoured in order to create
the pod config with the right number of vCPUs and Memory,
whithout this patch vCPU and Memory will be always 0.

fixes #925

### DIFF
--- a/config.go
+++ b/config.go
@@ -314,6 +314,9 @@ func updateRuntimeConfig(configPath string, tomlConf tomlConfig, config *oci.Run
 				return fmt.Errorf("%v: %v", configPath, err)
 			}
 
+			config.VMConfig.VCPUs = uint(hConfig.DefaultVCPUs)
+			config.VMConfig.Memory = uint(hConfig.DefaultMemSz)
+
 			config.HypervisorConfig = hConfig
 		}
 	}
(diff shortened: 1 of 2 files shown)


jodh-intel:
Looks good, but could you add some tests to `config_test.go` for this change?

jodh-intel:
Ugh - the CI's are complaining that docker isn't installed/running. Any ideas @chavafg?

devimc:
@jodh-intel changes applied, thanks

PR #931, opened 2018-01-16
CI: do not call .ci/setup.sh with sudo

### LINKED ISSUE #930: do not run CI setup as root
# Description of problem

Currently, the `.ci/setup.sh` from the tests repo is being called with sudo. There is no need to do it that way and it may make some things not work correctly when running with sudo.

chavafg:
There is no need to setup the CC environment with the
root user.

Also after running the setup, we needed to change the ownership
of the $GOPATH directory back from root. With this change we
will not need to do this anymore.

Fixes #930.

Signed-off-by: Salvador Fuentes <salvador.fuentes@intel.com>

### DIFF
--- a/.ci/run.sh
+++ b/.ci/run.sh
@@ -18,11 +18,6 @@
 
 set -e
 
-# Make sure root does not own any folder of the GOPATH
-# Otherwise this would prevent the following scripts to
-# succeed when performing "go get"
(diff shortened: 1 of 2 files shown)


jodh-intel:
lgtm

devimc:
lgtm

Issue #921, opened 2018-01-12
main: Remove hardcoded Clear Containers references


sboeuf:
Every reference to Clear Containers should be configured depending on the configuration used. In case we test Kata, we want the logger name to be `kata-runtime` instead of `cc-runtime`. Same thing for references to `Clear Containers`.
Refer to:
https://github.com/clearcontainers/runtime/blob/master/main.go#L33-L34

jodh-intel:
Agreed. The most flexible approach would be to define them in the `Makefile`:

```
PROJECT_NAME = Intel® Clear Containers
PROJECT_BIN_NAME = cc-runtime

USER_VARS += PROJECT_NAME
USER_VARS += PROJECT_BIN_NAME
```

And then add the appropriate code to `GENERATED_CODE` in `Makefile` so that `config-generated.go` contains these values, rather than defining them in `main.go`.

PR #933, opened 2018-01-17
build: Allow for kata project build

### LINKED ISSUE #919: makefile: Allow for kata build
I think we should add the possibility to generate everything for a `kata-runtime` binary from the Makefile. We would use `make kata` in order to generate everything we need.

jodh-intel:
Rework the build system to allow the runtime to be built for either
of the following:

- Clear Containers (`cc-runtime` binary and configuration)
- Kata Containers (`kata-runtime` binary and configuration)

The existing build rules are retained but there are now two new rules:

```
$ make build-kata-system
$ sudo make install-kata-system
```

Two new variables have also been added:

- `KATA_SYSTEM_BUILD`: set to any value to build/install for Kata.
- `SYSTEM_BUILD_TYPE`: build for either Clear Containers (set to `cc`) or
Kata (set to `kata`).

See `make help` for further details.

Note that configuration files are installed to a project-specific
directory, allowing for systems to be installed for both Kata and Clear
Containers.

Finally, the bash completion and collection scripts have been updated
to work for either project.

Fixes #919, #921.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,4 +1,4 @@
-# Copyright (c) 2017 Intel Corporation
+# Copyright (c) 2017-2018 Intel Corporation
 #
 # Licensed under the Apache License, Version 2.0 (the "License");
 # you may not use this file except in compliance with the License.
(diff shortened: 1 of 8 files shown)


devimc:
good patch!
lgtm

jodh-intel:
Could I get another review on this PR please?

jodh-intel:
ping @sboeuf, @sameo.

Makefile
```diff
@@ -21,31 +21,114 @@ for file in /etc/os-release /usr/lib/os-release; do \
...
+
+# If this environment variable is set to any value,
+# enable the Kata Containers system build.
+ifneq (,$(KATA_SYSTEM_BUILD))
+    system_build_type = $(KATA_TYPE)
+endif
+
+ifneq ($(SYSTEM_BUILD_TYPE),)
+    ifeq ($(SYSTEM_BUILD_TYPE),$(CC_TYPE))
+        system_build_type = $(SYSTEM_BUILD_TYPE)
```
grahamwhaley: Although equivalent, I think it looks less confusing to do:
```
system_build_type = $(CC_TYPE)
```
here, and the equiv for KATA below.

I guess all we are really doing here is checking that the `SYSTEM_BUILD_TYPE` is actually a type we know about. In which case, would an error case catchall of something like `ifeq ($(SYSTEM_BUILD_TYPE),)` after these two cases make sense? (that is, nobody set the type?) But, maybe that does not mesh with the other methods of choosing which sort of build (I sort of wish we only had one method....)

Makefile
```diff
@@ -21,31 +21,114 @@ for file in /etc/os-release /usr/lib/os-release; do \
     fi \
 done)
 
-# A "CC system build" is either triggered by specifying
-# a special target or via an environment variable.
+#------------------------------
```
grahamwhaley: Update copyright year in this file? - and check others?

jodh-intel:
Hi @grahamwhaley - branch updated.

sameo:
LGTM

sboeuf:
@jodh-intel FWIW, I have been testing Kata Containers with this PR (now in master), and I confirm it works very well as expected :)

PR #938, opened 2018-01-18
build: Add basic architecture framework

### LINKED ISSUE #928: Allow for builds on different architectures
Update the build system to allow for possible future architectures:

```
$ make ARCH=... ...
```

The default architecture will be the architecture of the host system the build is run on (currently assumed to be and limited to a `GOARCH` of `amd64`).

jodh-intel:
Support building on different architectures. Currently, only `amd64`
(`x86-64`) is supported. To add support for a new architecture, create:

```
arch/$(GOARCH)-options.mk
```

To build for a different architecture to the hosts, specify the `ARCH`
variable. For example:

```
$ make ARCH=... generate-config
```

Fixes #928.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -21,6 +21,19 @@ for file in /etc/os-release /usr/lib/os-release; do \
     fi \
 done)
 
+GOARCH=$(shell go env GOARCH)
+HOST_ARCH=$(shell arch)
+
+ifeq ($(ARCH),)
+	ARCH = $(GOARCH)
+endif
+
+ARCH_DIR = arch
+ARCH_FILE = $(ARCH_DIR)/$(ARCH)-options.mk
+
+# Load architecture-dependent settings
+include $(ARCH_FILE)
+
 #------------------------------
 # project-specifics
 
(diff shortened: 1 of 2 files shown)


jodh-intel:
ping @sameo, @sboeuf, @mcastelino, @grahamwhaley, @devimc.

jodh-intel:
Any comments folk?

jodh-intel:
re-ping @sameo, @sboeuf, @mcastelino, @grahamwhaley, @devimc ... :smile:

arch/amd64-options.mk
```diff
@@ -0,0 +1,27 @@
...
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+# Intel x86-64 settings
+
+MACHINETYPE := pc
```
devimc: not sure, but I think this should be inside an if sentence

```make
ifeq ($(ARCH),x86_64)
   MACHINETYPE := pc
endif
```

jodh-intel: I think we should be protected there from the check in the top-level `Makefile` ?

arch/amd64-options.mk
```diff
@@ -0,0 +1,27 @@
+# Copyright (c) 2018 Intel Corporation
```
devimc: ohh, I seee, this is amd64 specific,
I wonder if an if sentence is needed in the makefile in order to include or not this file

devimc:
@jodh-intel sorry for delay in my review
in general this PR lgtm

arch/amd64-options.mk
jodh-intel: That's a fair point - branch updated with a comment ;)

devimc:
LGTM

Makefile
```diff
@@ -21,6 +21,19 @@ for file in /etc/os-release /usr/lib/os-release; do \
     fi \
 done)
 
+GOARCH=$(shell go env GOARCH)
+HOST_ARCH=$(shell arch)
+
+ifeq ($(ARCH),)
+	ARCH = $(GOARCH)
+endif
```
grahamwhaley: just fmi, do these changes implement cross compilation, or do we always expect a native compile?
For cross-compilation, I was sort of expecting to see something like:
```
else
 if ARCH != GOARCH
   GOARCH = ARCH
```

to pass the requested `ARCH` (if different from the host) on to the golang build. Maybe I'm missing something though?

jodh-intel: These changes don't take any account of cross-compiling. The only reason I added `arch` output is to show the "real" host architecture along with the standard golang name for that same arch.

If we want to do cross-compiling, that's a bigger topic and should be handled on a different PR imho.

jodh-intel:
Coveralls report: https://coveralls.io/jobs/33040666

Merging...

PR #956, opened 2018-01-24
build: Set systemd unit name based on project

### LINKED ISSUE #955: Set systemd unit name based on project
The systemd unit name used to launch the agent in the guest needs to be set based on the project being built for.

jodh-intel:
Set the name of the systemd(1) unit based on the project being built for.

The unit name is used to construct the kernel command line, and then
used by the guests systemd to launch the agent, but previously only the
clear containers agent was supported.

Fixes #955.


sameo:
LGTM

PR #962, opened 2018-01-25
usage: Show project URL in help output

### LINKED ISSUE #961: Show project URL in help output
Display the project URL in the `--help` output.

This will be https://github.com/clearcontainers, regardless of which project the Clear Containers runtime is built for.

jodh-intel:
Display the project URL in the `--help` output which is fixed,
regardless of which project the Clear Containers runtime is built for.

Fixes #961.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -340,6 +340,9 @@ const projectPrefix = "$(PROJECT_TYPE)"
 // systemdUnitName is the systemd(1) target used to launch the agent.
 const systemdUnitName = "$(PROJECT_TAG).target"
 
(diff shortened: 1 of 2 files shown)


jodh-intel:
As mentioned https://github.com/kata-containers/runtime/issues/15#issuecomment-360425095, we need a way to distinguish the runtimes clearly.

Rather than adding magic tags to the version string, this PR just adds the Clear Containers URL to the help output even if you build as `make build-kata-system` so the source of the runtime is clear to users.

PR #972, opened 2018-01-30
scripts: Fix collect script bug URL

### LINKED ISSUE #971: Fix bug url in collect script
The URL shown when the user runs `{cc|kata}-collect-data.sh -h` is incorrect.

jodh-intel:
Correct the URL displayed when the user runs
`{cc|kata}-collect-data.sh -h`.

Fixes #971.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -554,7 +554,9 @@ show-header:
 	@printf "%s - version %s (commit %s)\n\n" $(TARGET) $(VERSION) $(COMMIT)
 
 show-footer:
-	@printf "• Project home: $(PROJECT_URL)\n\n"
+	@printf "• Project:\n"
+	@printf "\tHome: $(PROJECT_URL)\n"
+	@printf "\tBugs: $(PROJECT_BUG_URL)\n\n"
 
 show-summary: show-header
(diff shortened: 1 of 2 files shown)


Makefile
```diff
@@ -43,13 +43,15 @@ CC_TYPE = cc
...
 CC_PROJECT_URL = https://github.com/clearcontainers
+CC_BUG_URL = $(CC_PROJECT_URL)/runtime/issues/new
 
 # build type for Kata Containers
 KATA_TYPE = kata
 
 KATA_PROJECT_NAME = Kata Containers
 KATA_PROJECT_TAG = kata-containers
 KATA_PROJECT_URL = https://github.com/kata-containers
+KATA_BUG_URL = $(KATA_PROJECT_URL)/kata-containers/issues/new
```
grahamwhaley: I guess that is in theory the best, most generic and neutral, URL we have. Shame 'orgs' don't have an Issues section on github.

jodh-intel: Yeah. I suspect we'll end up with a bunch of issues both at this URL and at the runtime one but from a triage perspective, it would be good if we could get them reported in the "main" repo as defined here.

PR #990, opened 2018-02-06
stop: Improve kill and delete semantic

sboeuf:
We should consider "kill" command as something more complex than only
passing some signal. Indeed, in case this signal intend to terminate
the container process, we should expect the container to be properly
cleaned up regarding the internal mounts of the rootfs and additional
devices hotplugged to the VM.

In case of CRI-O, the caller expects to be able to unmount container
rootfs after the kill returns and after it has validated the process
is no longer running.

This patch covers this case, leaving more responsibility to the kill
command, which means that we expect a simple call to DeletePod() or
DeleteContainer() in case of "delete" command.

There is one more case which is the "delete --force" command, which
is simply handled as a "kill + delete" command.

Signed-off-by: Sebastien Boeuf <sebastien.boeuf@intel.com>

### DIFF
--- a/delete.go
+++ b/delete.go
@@ -18,6 +18,7 @@ package main
 import (
 	"fmt"
 	"os"
+	"syscall"
 
 	vc "github.com/containers/virtcontainers"
 	"github.com/containers/virtcontainers/pkg/oci"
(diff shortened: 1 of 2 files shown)


sboeuf:
Replaced by #1003

PR #1003, opened 2018-02-12
kill: Fix CRI-O race condition

### LINKED ISSUE #1002: kill: CRI-O race condition
As explained [here](https://github.com/containers/virtcontainers/issues/606), we have a race condition about the way a container is stopped with CRI-O, and actually this should be solved as this is the way any user of an OCI compliant runtime should think of the way a runtime works.
In short, using `cc-runtime kill SIGKILL|SIGTERM` to a container means that we expect this container not to have any remaining dependencies after its process returned.

sboeuf:
CRI-O assumes that anything related to the container should be able
to be unmounted after the container process returned after a kill.
In our case, we have a strong dependency on what should be done before
CRI-O can actually unmount storage because it is plugged into the VM.

This means we have to translate a kill as a stop when needed. The
condition for that is to detect when someone calling into kill would
actually expect to see the container completely stopped. This case
happens for SIGKILL and SIGTERM, and those are the conditions for
our runtime to consider stopping the container after it has been
signalled.

Fixes #1002

Signed-off-by: Sebastien Boeuf <sebastien.boeuf@intel.com>

### DIFF
--- a/Gopkg.lock
+++ b/Gopkg.lock
@@ -50,7 +50,7 @@
     "pkg/uuid",
     "pkg/vcMock"
   ]
-  revision = "f8d2c11dcfd5c04e529e5900a195ee055734d5df"
+  revision = "6621b5d57f99532d6b3a52208a79f63ceaa41308"
 
 [[projects]]
   name = "github.com/davecgh/go-spew"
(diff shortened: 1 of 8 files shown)


jodh-intel:
The change looks fine but could you add some unit tests (probably based on `TestKillCLIFunctionSuccessful`) that assert that `StopContainer()` is called if the signal isn't `SIGTERM`/`SIGKILL` and one to assert that it isn't if the signal is one of those two using ` testingImpl.StopContainerFunc = ...`?

sboeuf:
@jodh-intel yes that's definitely the plan, I wanted to make sure CRI-O is passing properly here first

sboeuf:
@jodh-intel I have fixed unit tests to reflect what we expect from a call to `kill`.

sameo:
LGTM

jodh-intel:
lgtm

PR #1018, opened 2018-02-21
logger: Fix nanosecond timestamp test

### LINKED ISSUE #1017: Fix nanosecond logging test
`TestNewSystemLogHook()` checks to ensure a nanonsecond timestamp is generated. However, `time.RFC3339Nano` truncates trailing zeros so the current test is incorrect.

jodh-intel:
`TestNewSystemLogHook()` checks to ensure a nanonsecond timestamp is
generated. However, the `time.RFC3339Nano` time format truncates
trailing zeros meaning a nanosecond timestamp won't always have nine
digits.

Fixes #1017.

### DIFF
--- a/logger_test.go
+++ b/logger_test.go
@@ -150,8 +150,10 @@ func TestNewSystemLogHook(t *testing.T) {
 			`\d{2}:\d{2}:\d{2}` +
 			// high-precision separator
 			`.` +
-			// nano-seconds
-			`\d{9}` +
+			// nano-seconds. Note that the quantifier range is
+			// required because the time.RFC3339Nano format
+			// trunctates trailing zeros.
+			`\d{1,9}` +
 			// UTC timezone specifier
 			`Z`
 


logger_test.go
```diff
@@ -150,8 +150,10 @@ func TestNewSystemLogHook(t *testing.T) {
 			`\d{2}:\d{2}:\d{2}` +
 			// high-precision separator
 			`.` +
-			// nano-seconds
-			`\d{9}` +
+			// nano-seconds. Note that the quantifier range is
+			// required because the time.RFC3339Nano format
+			// trunctates trailing zeros.
+			`\d{1,9}` +
```
grahamwhaley: just checking - I presume '0' comes out as '0', and hence always at least 1 digit? :-)

devimc: hehe, same question

devimc:
lgtm

logger_test.go
jodh-intel: That's a good question. It isn't documented afaics but observationally there is always a leading zero, yes. I've also discovered something rather interesting by looking at the parsing code:

- https://github.com/golang/go/blob/master/src/time/format.go#L252..L269

`time.RFC3339Nano` truncates trailing zeros because it is defined like this:

```
RFC3339Nano = "2006-01-02T15:04:05.999999999Z07:00"
```

But, you can create a non-truncating format by switching the values after the decimal point to zeros!:

```
const RFC3339NanoNoTruncate = "2006-01-02T15:04:05.000000000Z07:00"
```

Kinda surprising someone didn't add that as a standard option given the documented sorting behaviour of `time.RFC3339Nano`.

/cc @markdryan.

jodh-intel:
The metrics CI is again stopping the button from going green :-(

Force-merging as everything else is happy and this is causing the new release on #1021 to fail CI.

grahamwhaley:
Just to note, the metrics CI failed due to a network/docker hub issue:
```
07:00:21 Get https://registry-1.docker.io/v2/library/nginx/manifests/sha256:600bff7fb36d7992512f8c07abd50aac08db8f17c94e3c83e47d53435a1a6f7c: net/http: TLS handshake timeout
07:00:21 ERROR: Failed to docker pull image nginx
```

PR #1025, opened 2018-02-23
logging: Add CLI command to logs

### LINKED ISSUE #1024: add command name to log entries
Since the runtime is called multiple times by the container manager, it would be useful if all log entries show the name of the CLI (sub-)command that was running at the time the message was generated.

jodh-intel:
Since the runtime is called multiple times by the container manager, add
the command-line sub-command name to all log entries to make it clear
which command was running when a particular log message was produced.

Note that CLI arguments are already logged once at runtime startup. This
change is more useful for the scenarios where only a log excerpt is
available (for example when running `cc-collect-data.sh` which only
extracts error logs).

Fixes #1024.

### DIFF
--- a/main.go
+++ b/main.go
@@ -202,6 +202,13 @@ func beforeSubcommands(context *cli.Context) error {
 
 	ignoreLogging := false
 
+	// Add the name of the sub-command to each log entry for easier
+	// debugging.
+	name := context.Args().First()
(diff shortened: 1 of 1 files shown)


jodh-intel:
@grahamwhaley - no it doesn't. That should be atleast partly possible in a follow-on PR.

devimc:
lgtm

PR #1041, opened 2018-02-28
config: set 1 vCPU as the default number of vCPUs

### LINKED ISSUE #1032: [Proposal] [RFC] Start VMs with 1 vCPU
# Description
By default the runtime starts the VM with the same number of physical cores, sometimes when the number of physical cores is not too big, the boot time and memory consumption is not a problem, but when the number of physical cores is big, the boot time and memory footprint increase [*]:

**time to create and destroy a container**
```md
+-----------------------+----------------+---------------------+
|  Container CPUs_Mem   |  time(s)       | Memory footprint(Mb)|
+=======================+================+=====================+
|  2cpus_2gb            | 1.7401         | 68.36               |
+-----------------------+----------------+---------------------+
|  88cpus_2gb           | 2.34915        | 166.008008          |
+-----------------------+----------------+---------------------+
```

Other point to consider is the fact that the runtime must have support to update container's resources dynamically, and the only way to achieve this is through QEMU hot-plug and cgroups.
The VM should start with the minimum number of resources (CPU and Memory) to allow the runtime hot plug and unplug resources as needed.

The difference between starting the container with 1 and 2 vCPUs is not considerable [+]

**time to create and destroy a container**
```md
+--------------------+----------------+
|  Container vCPUs   |  time(s)       |
+====================+================+
|  1                 | 1.017          |
+--------------------+----------------+
|  2                 | 1.0129         |
+--------------------+----------------+
```

[*] thanks to @grahamwhaley  for the data
[+] using `launch_times.sh -i debian -n 100 -r cc-runtime`

# Proposal
- [ ] Change `default_vcpus` to 1 in the runtime configuration file 

Currently when a container with CPU constrains is created using docker, the VM starts with a different number of vcpus than the value specified by `default_vcpus`, this was an expected behaviour but with https://github.com/containers/virtcontainers/pull/512 a CPU cgroup is created inside the VM limiting the CPU usage of the container, hence doesn't matter if the VM has 89 vCPUs the container only can use the CPUs of the cgroup.

- [x] Always start the VM with the number of vCPUs specified by `default_vcpus`

devimc:
In order to have a fast boot time and small memory foot print,
this patch changes the default number of vCPUs to 1.
Other benefit of these change is the control over almost all the vCPUs,
all vCPUs less one will be hot pluggable.

fixes #1032

### DIFF
--- a/docs/cpu-constraints.md
+++ b/docs/cpu-constraints.md
@@ -1,0 +1,40 @@
+# CPU Constraints in Clear Containers
+
+## Default number of virtual CPUs
+
+Before starting a container, the runtime reads the option `default_vcpus` from the configuration file to determine the number of virtual CPUs (vCPUs) needed to start the virtual machine. By default, `default_vcpus` is equal to 1 for fast boot time and a small memory footprint per virtual machine. Be aware that increasing this value negatively impacts the virtual machine's boot time and memory footprint.
+In general, we recommend that you do not edit this variable, unless you know what are you doing. If your container needs more than one vCPU, use [docker `--cpus`][1], [docker update][4] or [kubernetes `cpu` limits][2] to assign more resources.
+
+## Virtual CPUs and Kubernetes PODs
+
+A kubernetes POD is a group of one or more containers, with shared storage and network, and a specification for how to run the containers [[1][3]]. In Clear Containers this group of containers runs inside the same virtual machine. If you do not specify a CPU constraint, the runtime does not hot add more vCPUs and the container is not placed inside a CPU cgroup.  Instead, the container uses the number of vCPUs specified by `default_vcpus` and shares these resources with other containers in the same situation (without a CPU constraint).
+
+## Containers life cycle
+
+When you create a container with a CPU constraint, the runtime hot adds the number of vCPUs required by the container. Similarly, when the container stops, the runtime hot removes these resources.
+
+## Containers without CPU constraint
+
+A container without a CPU constraint uses the default number of vCPUs specified in the configuration file. In the case of Kubernetes PODs, containers without a CPU constraint use and share between them the default number vCPUs. For example, if default_vcpus is equal to 1 and you have 2 containers without CPU constraints with each container trying to consume 100% of vCPU, the resources divide in two parts, 50% of vCPU for each container because your virtual machine does not have enough resources to satisfy containers needs. If you wish to give access to a greater or lesser portion of vCPUs to a specific container, you can use [docker --cpu-shares][1] or [kubernetes `cpu` requests][2].
+Before running containers without CPU constraint, consider that your containers are not running alone. Since your containers run inside a virtual machine other processes try to use the vCPUs as well (e.g. `systemd` and the Clear Container [agent][5]). In general, we recommend setting `default_vcpus` equal to 1 to allow non-container processes to run on this vCPU and to specify a CPU constraint for each container. If your container is already running and needs more vCPUs, you can add more using [docker update][4].
+
+## Containers with CPU constraint
+
+The runtime calculates the number of vCPUs required by containers with CPU constraint using the following formula: `(quota + (period -1)) / period`. The result determines the number of vCPU to hot plug into the virtual machine. Once the vCPUs have been hot added, the [agent][5] places the container inside a CPU cgroup. This placement allows the container to use only its assigned resources.
+
+## Do not waste resources
+
+If you already know the number of vCPUs needed for each container and POD, or just want to run them with the same number of vCPUs, you can specify that number using the `default_vcpus` option in the configuration file, each virtual machine starts with that number of vCPUs. One limitation of this approach is that these vCPUs cannot be hot removed later and you might be wasting resources. For example, if you set `default_vcpus` to 8 and run only one container with a CPU constrain of 1 vCPUs, you might be wasting 7 vCPUs since the virtual machine starts with 8 vCPUs and 1 vCPUs is hot added and assigned to the container. Non-container processes might be able to use 8 vCPUs but they use a maximum 1 vCPU, hence 7 vCPUs might not be used.
+
+## Limitations
+
+- [docker `--cpuset-cpus`][1], [docker --cpu-shares][1], [docker update][4] and [kubernetes `cpu` requests][2] are not supported.
+
+
+
+[1]: https://docs.docker.com/config/containers/resource_constraints/#cpu
+[2]: https://kubernetes.io/docs/tasks/configure-pod-container/assign-cpu-resource
+[3]: https://kubernetes.io/docs/concepts/workloads/pods/pod/
+[4]: https://docs.docker.com/engine/reference/commandline/update/
+[5]: https://github.com/clearcontainers/agent
+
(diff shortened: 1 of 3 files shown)


jcvenegas:
lgtm

sboeuf:
@klynnrif PTAL

docs/cpu-constraints.md
```diff
@@ -0,0 +1,28 @@
...
+
+## Virtual CPUs and Kubernetes PODs
+
+A kubernetes POD is a group of one or more containers, with shared storage/network, and a specification for how to run the containers [[1][3]]. In Clear Containers this group of containers run inside the same virtual machine, when a new container is created with a CPU constraint, using CPU quota and period the runtime calculates how many vCPUs are required by the container, the formula to calculate this is: `(quota + (period - 1)) / period`. If a CPU constraint is not specified, the runtime will not hot add more vCPUs and the container will not be placed inside a CPU cgroup, it will use the number of vCPUs specified by `default_vcpus` and share these resources with other containers in the same situation (without CPU constraint).
+
+
+## Container life cycle
+
+As was mention above, when a container is created with a CPU constraint, the runtime will hot add the number of vCPUs required by the container, in the same way, when it is stopped the runtime will hot remove these resources.
```
klynnrif: Suggested rewrite: When you create a container with a CPU constraint, the runtime hot adds the number of vCPUs required by the container. Similarly, when the container stops, the runtime hot removes these resources.

docs/cpu-constraints.md
```diff
@@ -0,0 +1,28 @@
...
+
+
+## Limitations
+
+- [docker `--cpuset-cpus`][1], [docker --cpu-shares][1] and [kubernetes `cpu` requests][2] are not honoured.
```
klynnrif: For consistency I suggest using "honored" instead of "honoured".

jodh-intel: Could you also make this a link to `docs/limitations.md` and update that doc with these limitations?

devimc:
@klynnrif @jodh-intel changes applied, thanks

jodh-intel:
lgtm.

@grahamwhaley may want to take a look at this one too though.

docs/cpu-constraints.md
```diff
@@ -0,0 +1,40 @@
...
+In general, we recommend that you do not edit this variable, unless you know what are you doing. If your container needs more than one vCPU, use [docker `--cpus`][1], [docker update][4] or [kubernetes `cpu` limits][2] to assign more resources.
+
+## Virtual CPUs and Kubernetes PODs
+
+A kubernetes POD is a group of one or more containers, with shared storage and network, and a specification for how to run the containers [[1][3]]. In Clear Containers this group of containers run inside the same virtual machine. If you do not specify a CPU constraint, the runtime does not hot add more vCPUs and the container is not placed inside a CPU cgroup.  Instead, the container uses the number of vCPUs specified by `default_vcpus` and shares these resources with other containers in the same situation (without a CPU constraint).
```
klynnrif: Subject/Verb disagreement with group/run - suggested edit: "In Clear Containers this group of containers runs inside the same virtual machine."

docs/cpu-constraints.md
```diff
@@ -0,0 +1,40 @@
...
+
+## Containers life cycle
+
+When you create a container with a CPU constraint, the runtime hot adds the number of vCPUs required by the container. Similarly, when the container stops, the runtime hot removes these resources.
+
+## Containers without CPU constraint
+
+A container without a CPU constraint uses the default number of vCPUs specified in the configuration file, in the case of Kubernetes PODs, containers without a CPU constraint use and share between them the default number of vCPUs, for example, if `default_vcpus` is equal to 1 and you have 2 containers without CPU constraint and each container tries to consume 100% of vCPU, since your virtual machine has not enough resources to satisfy containers needs, the resources are divided in two parts, 50% of vCPU for each container. If you wish to give access to a greater or lesser portion of vCPUs to a specific container, you can use [docker --cpu-shares][1] or [kubernetes `cpu` requests][2].
```
klynnrif: Suggested rewrite to break up this sentence: "A container without a CPU constraint uses the default number of vCPUs specified in the configuration file. In the case of Kubernetes PODs, containers without a CPU constraint use and share between them the default number vCPUs. For example, if `default_vcpus` is equal to 1 and you have 2 containers without CPU constraints with each container trying to consume 100% of vCPU, the resources divide in two parts, 50% of vCPU for each container because your virtual machine does not have enough resources to satisfy containers needs."

docs/cpu-constraints.md
```diff
@@ -0,0 +1,40 @@
...
+One thing to consider before running containers without CPU constraint is that your containers are not running alone, since your containers run inside a virtual machine, there other processes trying to use the vCPUs too, for example systemd and the Clear Container [agent][5]. In general, we recommend to set `default_vcpus` to 1 to allow non-container processes run on this vCPU and specify a CPU constraint for each container, if your container is already running and it needs more vCPUs, you can add more using [docker update][4].
```
klynnrif: Suggested rewrite to break up the sentences: "Before running containers without CPU constraint, consider that your containers are not running alone. Since your containers run inside a virtual machine other processes try to use the vCPUs as well (e.g. `systemd` and the Clear Container [agent][5]). In general, we recommend setting `default_vcpus` equal to 1 to allow non-container processes to run on this vCPU and to specify a CPU constraint for each container. If your container is already running and needs more vCPUs, you can add more using [docker update][4]."

docs/cpu-constraints.md
```diff
@@ -0,0 +1,40 @@
...
+
+## Containers with CPU constraint
+
+The runtime calculates the number of vCPUs required by containers with CPU constraint using the following formula: `(quota + (period -1)) / period`, the result determinate the number of vCPU to hot plug into the virtual machine. Once the vCPUs have been hot added, the [agent][5] places the container inside a CPU cgroup, allowing to the container use only its resources assigned.
```
klynnrif: Suggested rewrite to break up the sentences: "The runtime calculates the number of vCPUs required by containers with CPU constraint using the following formula: `(quota + (period -1)) / period`. The result determines the number of vCPU to hot plug into the virtual machine. Once the vCPUs have been hot added, the [agent][5] places the container inside a CPU cgroup. This placement allows the container to use only its assigned resources."

docs/cpu-constraints.md
```diff
@@ -0,0 +1,40 @@
...
+
+## Do not waste resources
+
+If you already know the number of vCPUs needed for each container and POD, or just want to run them with the same number of vCPUs, you can specify that number using the `default_vcpus` option in the configuration file, each virtual machine starts with that number of vCPUs, one limitation of this approach is that these vCPUs can not be hot removed later and you might be wasting resources, for example if you set `default_vcpus` to 8 and run only one container with a CPU constrain of 1 vCPUs, you might be wasting 7 vCPUs, since the virtual machine starts with 8 vCPUs and 1 vCPUs is hot added and assigned to the container, non-container processes might be able to use 8 vCPUs, but they use a maximum 1 vCPU, hence 7 vCPUs might not be used.
```
klynnrif: Suggested rewrite to break up the sentences: "One limitation of this approach is that these vCPUs cannot be hot removed later and you might be wasting resources. For example, if you set `default_vcpus` to 8 and run only one container with a CPU constrain of 1 vCPUs, you might be wasting 7 vCPUs since the virtual machine starts with 8 vCPUs and 1 vCPUs is hot added and assigned to the container. Non-container processes might be able to use 8 vCPUs but they use a maximum 1 vCPU, hence 7 vCPUs might not be used."

devimc:
@klynnrif  changes applied, thanks

PR #1057, opened 2018-03-09
scripts: Add agent version to collect script

### LINKED ISSUE #1055: Find a way to display the agent version
As mentioned https://github.com/clearcontainers/tests/issues/944#issuecomment-371157669, the runtime does *not* display the agent version, although the output of `cc-env` suggests it should/tries to:

```
$ cc-runtime cc-env|grep -A 2 Agent
[Agent]
  Type = "hyperstart"
  Version = "<<unknown>>"
```

The best strategy would seem to be to bind-mount the image (using `losetup`+`mount`) and run `cc-agent --version` now that https://github.com/clearcontainers/agent/pull/213 has landed.

The question though is where we should run this:

1. In the runtime

   This would allow the `cc-env` output to display the correct `Agent.Version=`.

1. In [`cc-collect-data.sh`](https://github.com/clearcontainers/runtime/blob/master/data/collect-data.sh.in)

   If we do this, we need to remove the `Agent.Version=` from the runtime.

Option (1) seems logical but is a bit... brace yourself for a very technical term... "*yucky*". That said:

- such code wouldn't be part of the main runtime lifecycle flows.
- we already have precedent for doing similar things as `cc-check` calls `modinfo(8)`.

wdyt @sboeuf, @sameo, @grahamwhaley, @mcastelino? Should we take a vote on this? :smile:

jodh-intel:
Add the agent version to the collect script output.

Fixes #1055.

### DIFF
--- a/data/collect-data.sh.in
+++ b/data/collect-data.sh.in
@@ -20,6 +20,8 @@ runtime=$(command -v "$runtime_name" 2>/dev/null)
 issue_url="@PROJECT_BUG_URL@"
 script_version="@VERSION@ (commit @COMMIT@)"
 
+typeset -r unknown="unknown"
+
 # Maximum number of errors to show for a single system component
 # (such as runtime or proxy).
 PROBLEM_LIMIT=${PROBLEM_LIMIT:-50}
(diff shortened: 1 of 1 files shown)


jodh-intel:
This will "work", but won't actually show the correct agent version until we get a new [image release](https://github.com/clearcontainers/osbuilder/releases) containing https://github.com/clearcontainers/agent/pull/213 (commit https://github.com/clearcontainers/agent/pull/213/commits/14c4a2d4c9ae4cc8a53e54f57ab6ba10a432fb80).

Tested with Kata too (run as `sudo kata-collect-data.sh`) which shows the following on my test box:

---

# Agent

version:
```
kata-agent version 0.0.1-33eecb2a445f906811a5bc9713d2dafd10768d18-dirty
```

---

sboeuf:
Thanks @jodh-intel for this PR.
LGTM

PR #1080, opened 2018-03-20
docs: Add some documentation details for VFIO

### LINKED ISSUE #1079: Add details regarding how devices passed with VFIO can be accessed with a container

amshinde:
Add explanation regarding how devices passed with VFIO can
be accessed with a container.

Fixes #1079

### DIFF
--- a/docs/Passing-Devices-with-Clear-Containers.md
+++ b/docs/Passing-Devices-with-Clear-Containers.md
@@ -29,6 +29,15 @@ to `vfio-pci` driver and then pass the IOMMU group that the device belongs to on
 the docker command line.
 For detailed steps to bind and unbind devices, see https://github.com/containers/virtcontainers#how-to-pass-a-device-using-vfio-passthrough
 
+You can verify that the device is passed to the Clear Containers container by running
+`lspci` command inside the container.
+
+For the device to be visible inside the container under /dev, you need to pass /dev
+while starting a container as shown below:
+
+```
+$ sudo docker run -it -v /dev:/dev --device=/dev/vfio/16 centos/tools bash
+```
 
 ## Block Devices
 


amshinde:
@egernst Can you take a look. I am planning to move this Kata eventually.

mcastelino (commented):
@amshinde we need to pass -v /dev:/dev so that the VM's /dev is passed into the container /dev right.

Do we filter out any of the /dev entries? Or for that matter just bind mount the /dev corresponding to the device to the container?

amshinde:
@mcastelino yes, we are  doing that so that devices created in the VM's /dev appear in the container /dev.  We filter out initially by mounting /dev as tmpfs so we have a very limited set of devices to begin with  and then use udev events to bind mount all devices that appear after the container has started. We dont have any filtering rules at present for the bind-mounting, these would be any devices that are created after the container has started mostly due to the user interacting on the container command line.

amshinde:
@mcastelino For Kata,  plan is to pass some predictable identifiers for the VFIO devices specifically so that we do not need to pass -v /dev.  I am looking at things such as vendor-id that can be leveraged to identify the VFIO device passed, so that the corresponding device node can be created inside the container. Still looking at this.

docs/Passing-Devices-with-Clear-Containers.md
```diff
@@ -29,6 +29,15 @@ to `vfio-pci` driver and then pass the IOMMU group that the device belongs to on
...
 For detailed steps to bind and unbind devices, see https://github.com/containers/virtcontainers#how-to-pass-a-device-using-vfio-passthrough
 
+You can verify that the device is passed to the Clear Containers container by running
+`lspci` command inside the container.
+
+For the device to be visible inside the container under /dev, you need to pass /dev
+while starting a container as shown below:
+
+```
+$ sudo docker run -it -v /dev:/dev --device=/dev/vfio/16 centos/tools bash
```
egernst: I think this should be combined with the initial CLI example, and then perhaps explained as you did above.  

There's no usefulness in passing it into the VM, but not the container, after all.

mcastelino: @amshinde per OCI spec, the devices are container specific resources. So this approach will work fine for docker containers. How do we plan to handle this for kubernetes PODs, where the devices may be spread across containers (i.e. each device belongs to a different container for example)

amshinde: @mcastelino Containers in a k8 pod should share the same mount namespace eventually.

